PR #56, opened 2019-05-17
Promote FakeDataAgent to 00 status

mhasself:
Caution: now that FakeDataAgent has moved from the miniobs example to the core agents list, we can truly say it has... a License to Thrill.


jlashner:
I always knew FakeDataAgent would go on to do great things! I'm so proud

Issue #60, opened 2019-06-21
Aggregator prov_id list sync issue


BrianJKoopman:
We've seen a similar bug to #36 where the aggregator loses track of a provider under certain circumstances and fills the logs with errors like:

```
2019-06-21T20:16:19+0000 OCSAgent.onUserError(): “While firing <bound method DataAggregator._data_handler of <__main__.DataAggregator object at 0x7fba72af7198>> subscribed under 6560288436609885.”
Traceback (most recent call last):
 File “/usr/local/lib/python3.6/dist-packages/autobahn/twisted/websocket.py”, line 322, in _onMessage
   self.onMessage(payload, isBinary)
 File “/usr/local/lib/python3.6/dist-packages/autobahn/wamp/websocket.py”, line 95, in onMessage
   self._session.onMessage(msg)
 File “/usr/local/lib/python3.6/dist-packages/autobahn/wamp/protocol.py”, line 726, in onMessage
   future = txaio.as_future(handler.fn, *invoke_args, **invoke_kwargs)
 File “/usr/local/lib/python3.6/dist-packages/txaio/tx.py”, line 429, in as_future
   return maybeDeferred(fun, *args, **kwargs)
--- <exception caught here> ---
 File “/usr/local/lib/python3.6/dist-packages/twisted/internet/defer.py”, line 151, in maybeDeferred
   result = f(*args, **kw)
 File “aggregator_agent.py”, line 168, in _data_handler
   prov_id = self.prov_ids[feed[“address”]]
builtins.KeyError: ‘observatory.LSA24MA.feeds.temperatures’
```

I don't know how to reproduce this yet, but typically it occurs when someone goes wrong with the Agent like in [socs Issue #6](https://github.com/simonsobs/socs/issues/6). The LSA24MA 240 Agent was restarted, but the aggregator still produced these errors.

Aggregator version: v0.5.0-29-g2265349

BrianJKoopman:
Alright, so we've figured out how to reproduce this bug. Start with a freshly started aggregator and 240 Agent. Then restart the 240 Agent via a `docker-compose restart <agent name>` command. What occurs is when the 240 Agent is stopped, the aggregator schedules the removal of the associated provider. Then the 240 Agent is restarted, the aggregator realizes the `prov_id` is different and replaces it. Then the scheduled removal occurs, removing the new instance of the provider with updated `prov_id`. Here's more complete log output:

```
[...]

Issue #61, opened 2019-07-01
Aggregator - Sometimes prov is None when removing provider


BrianJKoopman:
Sometimes when the aggregator goes to remove a provider it finds `prov` to be `None`. This example occurred when a restarted a Lakeshore 240 Agent. But it doesn't happen every time I restart an agent, which makes me think some sort of race condition is occurring.

```
2019-07-01T21:52:08+0000 Added provider observatory.LSA22Z2.feeds.temperatures (agent sess: 1562017928.4343128) with id 2
2019-07-01T21:52:08+0000 Feed observatory.LSA22Z2.feeds.temperatures is already subscribed to
2019-07-01T21:52:09+0000 Removing provider observatory.LSA22Z2.feeds.temperatures with agent id 1562017600.4147737
2019-07-01T21:53:13+0000 OCSAgent.onUserError(): "While firing <bound method DataAggregator._new_agent_handler of <__main__.DataAggregator object at 0x7f93c3ece2b0>> subscribed under 7891720364125154."
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/autobahn/twisted/websocket.py", line 322, in _onMessage
    self.onMessage(payload, isBinary)
  File "/usr/local/lib/python3.6/dist-packages/autobahn/wamp/websocket.py", line 95, in onMessage
    self._session.onMessage(msg)
  File "/usr/local/lib/python3.6/dist-packages/autobahn/wamp/protocol.py", line 726, in onMessage
    future = txaio.as_future(handler.fn, *invoke_args, **invoke_kwargs)
  File "/usr/local/lib/python3.6/dist-packages/txaio/tx.py", line 429, in as_future
    return maybeDeferred(fun, *args, **kwargs)
--- <exception caught here> ---
  File "/usr/local/lib/python3.6/dist-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "aggregator_agent.py", line 209, in _new_agent_handler
    if action == 'removed' and prov.agent_id == agent_id:
builtins.AttributeError: 'NoneType' object has no attribute 'agent_id'

2019-07-01T21:53:13+0000 Added provider observatory.LSA22Z2.feeds.temperatures (agent sess: 1562017993.426738) with id 3
```

The relevant lines in the aggregator for context:
```
209             if action == 'removed' and prov.agent_id == agent_id:
210                 self.log.info("Scheduled remove for provider {}"
[...]

PR #62, opened 2019-07-03
Fix agg prov_id sync issue

jlashner:
This PR should fix issues #60 by changing the keys of the aggregator's prov_id dict from `feed_address` to `(feed_address, agent_session_id)`. This way the aggregator can avoid accidentally removing a new provider instead of an old one. I think it also fixe issue #61.

It fixed docker-compose restart with the fake-data-agent. Brian can you check if this fixes the problem with the Lakeshore agent? If it works I'll merge it.

### DIFF
--- a/agents/aggregator/Dockerfile
+++ b/agents/aggregator/Dockerfile
@@ -11,6 +11,9 @@ WORKDIR /app/agents/aggregator/
 RUN mkdir -p /data && \
     chown ocs:ocs /data
 
+# Copy this local copy into place over the OCS base image
+COPY aggregator_agent.py /app/agents/aggregator/
+
 # Run registry on container startup
 ENTRYPOINT ["python3", "-u", "aggregator_agent.py"]
 
--- a/agents/aggregator/aggregator_agent.py
+++ b/agents/aggregator/aggregator_agent.py
@@ -427,8 +427,11 @@ class DataAggregator:
                               .format(prov.address, prov.agent_id))
                 pid = prov.prov_id
                 self.hksess.remove_provider(pid)
-                del self.prov_ids[prov.address]
                 del self.providers[pid]
+
+                if self.prov_ids[prov.address] == pid:
+                    del self.prov_ids[prov.address]
+
 
             # Then write status if we need to
             if self.should_write_status:
--- a/agents/aggregator/aggregator_agent.py
+++ b/agents/aggregator/aggregator_agent.py
@@ -142,7 +142,7 @@ class DataAggregator:
         self.data_dir = data_dir
 
         self.providers = {} # by prov_id
-        self.prov_ids = {} # by provider address
+        self.prov_ids = {} # by (address, session_id)
         self.next_prov_id = 0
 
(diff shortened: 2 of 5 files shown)


BrianJKoopman:
I haven't had any coffee yet today. I realized I just responded on the issue, but meant to respond here...

> Great, thanks for the patch. I just tested it with one of our 240s and it seems to work.
> 
> My one comment (which is necessarily related to the bug, but while we're here fixing related thing...) is that we should use a uniform name for 'agent_session_id', the logs refer to 'agent sess' and 'agent id'. It seems both are an abbreviation of 'agention_session_id', but we should be consistent, otherwise the casual observer might not know they're actually the same thing.
> 
> Please address that, then merge away.

BrianJKoopman:
Actually, we've had students using the new aggregator here and have found that the problem still persists. It isn't evident until you do the previous set of steps, then command the 240 to acquire data. Once you tell it to acquire data the old errors return:

```
2019-07-03T18:12:53+0000 OCSAgent.onUserError(): "While firing <bound method DataAggregator._data_handler of <__main__.DataAggregator object at 0x7f23bc7f5390>> subscribed under 1145008383491662."                                                                                                                         
  File "/usr/local/lib/python3.6/dist-packages/autobahn/twisted/websocket.py", line 322, in _onMessage
    self.onMessage(payload, isBinary)
  File "/usr/local/lib/python3.6/dist-packages/autobahn/wamp/websocket.py", line 95, in onMessage
    self._session.onMessage(msg)
  File "/usr/local/lib/python3.6/dist-packages/autobahn/wamp/protocol.py", line 726, in onMessage
    future = txaio.as_future(handler.fn, *invoke_args, **invoke_kwargs)
  File "/usr/local/lib/python3.6/dist-packages/txaio/tx.py", line 429, in as_future
    return maybeDeferred(fun, *args, **kwargs)
--- <exception caught here> ---
  File "/usr/local/lib/python3.6/dist-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "aggregator_agent.py", line 168, in _data_handler
    prov_id = self.prov_ids[feed["address"]]
builtins.KeyError: 'observatory.LSA24MA.feeds.temperatures'
```

jlashner:
Are you sure that's on the right branch? Line 168 in this PR looks different from the code in the error message.

BrianJKoopman:
Ah, good catch. I did rebuild on the correct branch, but the Dockerfile doesn't copy in the `aggregator_agent.py`, it relies on the one in the ocs base image, which I hadn't rebuilt. I'm going to push a change to this branch with an update to the Dockerfile which copies it, as I think it should do that.

jlashner:
You can also mount the local copy of ocs into your docker by adding 
```      
- /path/to/ocs:/app
```
to your volumes. That's what I've been doing.

jlashner:
Ok I changed all instances of `agent_session_id` to just `session_id` because it makes more sense. The only reason it wasn't that before was because there was also the Operation session_id which was conflicting.

PR #63, opened 2019-07-24
Fix Aggregator timestamping

jlashner:
This PR just makes the aggregator use time.time() instead of datetime.utcnow() for file names. It shouldn't change anything in the data, but the way I was doing it before, getting the timestamp from the datetime object, is definitely incorrect and may have been causing some discrepancy in file names or start times.

### DIFF
--- a/agents/aggregator/aggregator_agent.py
+++ b/agents/aggregator/aggregator_agent.py
@@ -1,5 +1,5 @@
 import time, threading
-from datetime import datetime
+import datetime
 import binascii
 import numpy as np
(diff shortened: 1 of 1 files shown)


BrianJKoopman:
Great, yeah, it looks like `datetime.utcnow()` is doing something with timezones?

```
>>> time.time()
1564152341.3921819
>>> datetime.utcnow().timestamp()
1564166742.221222
```

Those differ by 4 hours, which is the difference currently of EST from UTC. I agree `time.time()` should be used.

PR #71, opened 2019-08-30
Fix log formatting error

BrianJKoopman:
This log message always produces an "Unable to format event" message in the
logs. This is due to the string format of a SessionDetails object containing
curly brackets. In the txaio logger curly brackets need to be escaped by
additional curly brackets. Here we implement this escape to properly format the
logs.

What exactly do we want to be logging from the SessionDetails? I'm wondering if this should be at a different log level.

### DIFF
--- a/ocs/ocs_agent.py
+++ b/ocs/ocs_agent.py
@@ -143,7 +143,7 @@ class OCSAgent(ApplicationSession):
 
     @inlineCallbacks
     def onJoin(self, details):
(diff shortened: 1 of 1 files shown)


BrianJKoopman:
Talked about this on the phone, something like this should work:
`self.log.info('session joined: {x}', x=details)`

BrianJKoopman:
Formatting fixed as suggested yesterday. Merging.

PR #73, opened 2019-09-30
Add bit to flip to turn agent heartbeat off

BrianJKoopman:
This adds an internal attribute for turning off an Agent's heartbeat. This is for reproducing an error state that was seen on one of the deployed test systems where an agent became unregistered with the registry, but was still running, causing the aggregator to raise an error.

We also add a debug log statement for showing the heartbeat itself, this is useful in numerous occasions, though is quite loud in the logs.


BrianJKoopman:
Coupled with https://github.com/simonsobs/socs/pull/26. Will merge by the end of the day unless there are objections.

PR #76, opened 2019-10-18
Adds some additional subscribe functionality to OCSAgent.

jlashner:
This PR adds some convenient subscribe functions to the OCSAgent. 

- It overrides the OCSAgent subscribe function to prevent re-subscription to topics by default.
This is important for the aggregator, and agents where the init function might be called multiple times.

- It adds a `subscribe_on_start` function, that tells the OCSAgent to immediately subscribe to 
a particular feed on join so you don't have to call it from an init task.

- Adds easier subscription options to `subscribe_to_feed`, so to do pattern matching, you can just add the argument `option={'match': 'wildcard'}` or `option={'match': 'prefix'}`

### DIFF
--- a/ocs/ocs_agent.py
+++ b/ocs/ocs_agent.py
@@ -470,8 +470,6 @@ class OCSAgent(ApplicationSession):
                 Feed name, e.g. `temperatures`
             handler (callable):
                 handler called with message data
-            topic (string):
-                uri of topic to subscribe to
             options (dict):
                 Dict or subscribe options. See https://autobahn.readthedocs.io/en/latest/reference/autobahn.wamp.html#autobahn.wamp.types.SubscribeOptions
             force_subscribe (bool):
--- a/ocs/ocs_agent.py
+++ b/ocs/ocs_agent.py
@@ -10,11 +10,10 @@ from twisted.internet.error import ReactorNotRunning
 from twisted.python import log
 from twisted.logger import formatEvent, FileLogObserver
 
-from autobahn.wamp.types import ComponentConfig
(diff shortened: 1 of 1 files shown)


ocs/ocs_agent.py
```diff
@@ -423,18 +427,72 @@ def publish_to_feed(self, feed_name, message, from_reactor=False):
...
+
+        Args:
+            agent_addr (str):
+                Full agent address, e.g. `observatory.LS12345`
+            feed_name (str):
+                Feed name, e.g. `temperatures`
+            handler (callable):
+                handler called with message data
+            topic (string):
+                uri of topic to subscribe to
```
mhasself: ^^^ Here; `topic` is not an arg.

BrianJKoopman:
Just a couple of questions so I make sure I'm following:
- What instances is forcing a re-subscription desired?
- What would occur if topics were subscribed to multiple times before this patch? (i.e. in the aggregator)

jlashner:
Usually it isn't desired, but I can imagine you might want to use `force` if an agent wants to subscribe to a topic with multiple different callbacks. I haven't ever wanted to use that, but since that's the default behavior of subscribe I thought I should include the option to do that.

Before this patch the re-subscription check was in the `subscribe_to_feed` function, so the aggregator still wouldn't have been able to subscribe to the same feed multiple times, but there was no check if you were to try to subscribe to an arbitrary topic multiple times. This just moves it to the `subscribe` function so that all subscriptions are checked instead of just OCS feeds. This also does not protect against subscribing to the same topic multiple times with different wildcard strings.

Issue #75, opened 2019-09-30
Agents are unaware of their registry status


BrianJKoopman:
## Issue
Agents are currently unaware of their registration status. If they become unregistered for some reason while they are still functional they do not know and thus do not attempt to reestablish their registration with the registry agent. This will currently cause any data they publish to not be saved the disk by the aggregator.

## Proposal
Agents should subscribe to the "agent_activity" feed and watch for themselves to be unregistered. If they are, for whatever reason, unregistered they should try to reestablish their registration with the registry agent.

## Related Issues
This is somewhat related to #67. The agents should be able to re-register themselves if the registry restarts. If the registry doesn't broadcast removals before restart then probably this will require a separate solution.

#74 is related in that the state that is caused by these two issues was observed by @Zhilei during testing.

mhasself:
If I may counter-propose, I think the feed subscription scheme might already be over-complicated, and we should take this opportunity to make a simpler, more robust system (rather than adding a layer to handle drops+reconnects).

Can we explore the possibility of discarding registration and heartbeats... and instead having Agents announce their complete registration information regularly, on a timer?  This will provide both a heartbeat and whatever information the aggregator / registry need to initialize agent or feed handling.  If anything dies, it need only listen for a moment to learn about everything that is running on the system.

BrianJKoopman:
Definitely! I wanted to open the issue for discussion.

I do agree that maybe things are a bit over-complicated right now. I think our goal in introducing the registry was to have a central authority for what agents were running. Currently the way an agent tells the registry about itself is by calling one of the registry agent's tasks. In doing so it describes itself with this info:
```
        return {
            'agent_address': self.agent_address,
            'session_id': self.agent_session_id,
            'feeds': [f[1].encoded() for f in self.feeds.items()],
            'tasks': list(self.tasks.keys()),
            'processes': list(self.processes.keys())
        }   
```

So we could publish this as the replacement heartbeat every X seconds and remove the registry altogether. We'd still have to somehow tell the aggregator what URI to subscribe to in order to listen for this info.

jlashner:
I'm not really here, but you can subscribe to feeds with wildcards, so anything with the feedname  `root.<instance>.heartbeat`

mhasself:
Currently we have to tell all agents the address of the registry (so they can call the registration tasks).  Perhaps we could retain that setting, but have agents all yell their status on some single feed in that namespace (``observatory.registry.feeds.agent_announce``?).

[...]

Issue #74, opened 2019-09-30
Aggregator remains subscribed to removed provider feeds


BrianJKoopman:
The OCS Aggregator remains subscribed to provider feeds, even if those providers have been removed by the registry.

If, for some reason, an Agent is unregistered (for instance, if the registry loses track of the agent heartbeat for more than 5 seconds), then the registry removes the agent from the registered agents list. The aggregator, in turn, removes it from the providers and prov_id lists. However, if the agent is still actually active and publishing data to the feed, then the aggregator still receives that data and the `_data_handler()` method is still invoked. This [method tries to reference the agent](https://github.com/simonsobs/ocs/blob/5835febc52e0ae65c3ef0f22a9bcb0f89bc9ccf1/agents/aggregator/aggregator_agent.py#L168) address and session ID as a key to the prov_id list. This raises a KeyError in the aggregator.

I think the best solution here is to have the aggregator unsubscribe from the feeds that are removed. This will require the OCSAgent class track its subscriptions in order to call `sub.unsubscribe()`. See [example from autobahn repo][1].

## Testing
This scenario can be tested with some new code introduced in https://github.com/simonsobs/ocs/pull/73, https://github.com/simonsobs/socs/pull/25, and https://github.com/simonsobs/socs/pull/26. 

Setup these containers:
* crossbar server
* registry
* aggregator
* ls240 agent
* ls240 simulator
* ocs client

You can observe the ls240 agent begin to take data, the registry register the agent, and the aggregator record the data. Then if you disable the ls240 agent heartbeat the described errors occur.

[1]: https://github.com/crossbario/autobahn-python/blob/aedf45b09058de18e1a7ae5d73ee868051e9458d/examples/twisted/wamp/pubsub/unsubscribe/frontend.py

mhasself:
I believe this is not a concern, after #78 .

Issue #67, opened 2019-08-01
Agent registration should be robust to registry agent restarts


BrianJKoopman:
Currently if the OCS Registry Agent is restarted then any previously registered Agents need to also be restarted. When working on a multi-node system this isn't immediately obvious. The agents will keep running, but no data will be aggregating from them.

mhasself:
Resolved in #78

PR #78, opened 2019-10-20
Aggregator 2.0

jlashner:
This is a refactoring of the HK Aggregator which I've been wanting to do for a while. The main goals of the rewrite are to:

 - Separate the aggregator logic from crossbar, and divide it into clearly defined and easily testable functions instead of having most of the logic being in the `run` task, so it is easier to debug.
- Change the way in which providers are handled. In this new version, the aggregator doesn't rely on the registry to get feed data. Instead it listens to all feeds using wildcard matching, and if they are "recorded feeds" it registers the provider automatically. If it doesn't get any data from the provider in some amount of time (right now its hardcoded to 3 min, but maybe it should be a feed option), the provider is marked as stale and removed. I think this is fine, since as soon as it starts sending data again, the aggregator will automatically re-register it, with the only difference being the HKSession internal provider id.

I wanted to add the tests in this PR, but I'm having trouble updating so3g to python3.7 on my laptop, so I'll need to dockerize the tests which I think might be out of the scope of this PR. Hopefully this will solve a lot of the registration problems that we are having (i.e. #75, #74, #67)

### DIFF
--- a/agents/aggregator/aggregator_agent.py
+++ b/agents/aggregator/aggregator_agent.py
@@ -1,14 +1,9 @@
-import time, threading
-import datetime
-import binascii
-import numpy as np
-from ocs import ocs_agent, site_config, client_t, ocs_feed
-import os
-
-from twisted.internet.defer import inlineCallbacks
-from twisted.internet import reactor
+import time, datetime, binascii, os, queue, argparse
 from threading import RLock
+from typing import Dict
 
+from ocs import ocs_agent, site_config, ocs_feed
+import argparse
 if os.getenv('OCS_DOC_BUILD') != 'True':
     from spt3g import core
     import so3g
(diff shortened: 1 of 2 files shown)


agents/aggregator/aggregator_agent.py
```diff
@@ -18,463 +13,511 @@ class Provider:
...
 
-        for pid, prov in self.providers.items():
-            if prov.frame_start_time is None:
-                continue
+        self.rotator.Process(self.hksess.session_frame())
 
-            with prov.lock:
-                frame = prov.to_frame(self.hksess)
-                prov.clear()
+        self.aggregate = True
```
mhasself: This should probably go at the very top of the function -- or at least before the Process() -- to reduce the race-condition against stop().

agents/aggregator/aggregator_agent.py
```diff
@@ -18,463 +13,511 @@ class Provider:
...
-                    with prov.lock:
-                        frame = prov.to_frame(self.hksess)
-                        prov.clear()
-                    self.writer(frame)
+        address = feed['address']
+        sessid = feed['session_id']
+        frame_length = feed['agg_params']['frame_length']
 
-            self.writer.Flush()
+        pid = self.aggregator.pids.get((address, sessid))
```
mhasself: This is an unsafe access to self.aggregator internal variables -- you should use a method call, because that can be inspected for thread-safety in the internal context of the Aggregator class.

It's likely that you'll need to protect Aggregator.pids / aggregator.providers with some kind of lock.

agents/aggregator/aggregator_agent.py
```diff
@@ -18,463 +13,511 @@ class Provider:
...
-                    (os.getpid(), 14),
-                    (binascii.crc32(bytes(self.hksess.description, 'utf8')), 14)]
-        agg_session_id = 0
-        for i, b in elements:
-            agg_session_id = (agg_session_id << b) | (i % (1<<b))
-        self.log.info("New aggregator session_id=%i" % agg_session_id)
-        # The cast here to G3Int can be dropped in the future -- spt3g
-        # was updated on May 1 2019 to handle direct puts of int64_t.
-        self.hksess.session_id = core.G3Int(agg_session_id)
+        # SUBSCRIBES TO ALL FEEDS!!!!
```
mhasself: Indeed, this is a significant change.  Have you taken a census of what Agents currently use non-recorded feeds?  I'm not against having separate domains for "recordable" and "ephemeral" data streams.  (Certainly for now it should be the ephemeral ones that move elsewhere.)

agents/aggregator/aggregator_agent.py
```diff
@@ -18,463 +13,511 @@ class Provider:
...
-
-        ``params`` is a dict with the following keys:
+        self.hksess.remove_provider(pid)
+        del self.providers[pid]
+        del self.pids[(addr, sessid)]
+        self.write_status()
 
-        - 'start_record' (bool, optional): Default is False.  If True,
-            start the "record" Process after performing registration.
+    def write_data(self, data, pid):
```
mhasself: Within this class, "write" is used to describe each of:
- transfering data into a Provider storage object.
- transfering data from Providers to our queue.
- transfering data from queue to file.

So maybe "write_data" should be more descriptive, like the other write methods -- "write_data_to_provider"?  My preference would be to use different verbs though... store/enqueue/write?

jlashner:
Awesome, thanks for the feedback. I'm having some trouble figuring out how to keep this threadsafe as long as providers can be written to from the feed callbacks, which I'm pretty sure happens in the reactor thread. I'm think the way I'm doing it right now is bad because the providers write function will block if the provider is in the middle of writing a frame.

Would it be better to restrict all provider access to the `run` worker thread? The data callback could just write the data and feed info to a threadsafe queue, and the actual work could happen in the run function.

Does anyone have any better suggestions?

mhasself:
Actually I think there's no huge blocking problem with your data path, currently.  The Provider object is _slightly_ blocking, because the lock could be held by either the reactor or the worker thread.  But everything that is done with the lock held is fast (no I/O).  So although such locks might violate the spirit of asynchronous frameworks... it's probably fine.  It's really no worse than using a Queue to pass objects from reactor to worker thread.

That being said, I'm not against your proposal of having all processing occur in the worker thread.  With a single queue to pass data from reactor to worker, you can probably do away with all the other locking (Provider._lock, Aggregator.frame_queue).

agents/aggregator/aggregator_agent.py
```diff
@@ -18,463 +13,511 @@ class Provider:
...
 
-        self.end_file()
-        return True, 'Acquisition exited cleanly.'
+        self.aggregator.write_data(data, pid)
+
+    def start_aggregate(self, session, params=None):
+        """
+        Process for starting data aggregation
+        """
+        self.aggregator.run()
```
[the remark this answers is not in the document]
jlashner: I thought that there was code built into the ocs_agent start function that prevents the same process from being run in separate threads... Is that not true?

jlashner:
Ok cool, I'm liking the idea of having a single `incoming_data` queue that is processed from the run function, and then getting rid of all other locks. Seems more reliable than just adding more locks for the Aggregator.pids and Aggregator.providers objects, along with the provider locks.

agents/aggregator/aggregator_agent.py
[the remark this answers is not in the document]
mhasself: Oh, I think you're right about the single invocation.  But you should still have
  session.set_status('running')
in the start function and
  session.set_status('stopping')
in the stop function.

agents/aggregator/aggregator_agent.py
jlashner: I think most other "ephemeral" feeds are heartbeats (its arguable whether they should fall under 'feeds' in the first place). I think currently, status messages don't actually match this subscription, because they're published to `<agent_address>.feed`. The registry has the `agent_activity` feed that's not recorded, but that doesn't really need to exist after this PR... Just did a quick search of ocs and socs, and it actually doesn't look like there are any other non-aggregated feeds.

agents/aggregator/aggregator_agent.py
```diff
@@ -18,463 +13,511 @@ class Provider:
...
+        if (time.time() - self.file_start_time) > self.time_per_file:
+            self.close_file()
 
 
-class DataAggregator:
+class Aggregator:
     """
-    The Data Aggregator Agent listens to data provider feeds, and outputs the housekeeping data as G3Frames.
+    The Data Aggregator Agent listens to data provider feeds, and outputs the
+    housekeeping data as G3Frames.
```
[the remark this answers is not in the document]
jlashner: Would it be less awkward if instead I made all the non-thread-safe variables (pids, providers, frame_queue, ...) local to the run function, and passed as arguments to the functions? Then there would be no question about if it is unsafe to call `add_provider`, but it would make it more awkward to call, since you'd have to pass in the `providers` and `pids` objects.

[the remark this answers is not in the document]
jlashner: Or maybe all of those objects and functions that should be put into a ProviderCollection class, that is instantiated in the run thread... That seems like the best way of doing it.

jlashner:
Ok I swear this is the last major restructuring unless there are any other serious problems, since I feel like I'm kind of going around in circles.... I decided to separate all of the non-threadsafe functionality and data management into the Aggregator class, and put all of the thread-safe operations (or at least the enqueue_incoming_data function) into the AggregatorAgent class, which is the WAMP wrapper. The Aggregator instance is created local to the `run` function when it starts, so hopefully there's no chance of threads accessing data they shouldn't. To pass data to the Aggregator, the feed handler writes to a thread-safe queue `incoming_data`, which is passed to the Aggregator during the run function. I think this addresses Matthew's comments about making it clear in the code which functions are thread-safe and which are not.

mhasself:
Thanks for re-organizing.  If I could make a few more suggestions (sorry):
- I think this is a good way to organize it... but I think most of the stuff in start_aggregate should still be in an Aggregator function.  (I guess you got rid of run() because it had a loop that was controlled by a variable manipulated through stop_aggregate().)  I am thinking that it would be better to have writer and hksess private to Aggregator, and then start_aggregate would look like this:
```
        session.set_status('running')
        self.log.info("Aggregator running...")
	self.aggregate = True
        aggregator = Aggregator(data_queue=self.incoming_data_queue)
        while self.aggregate:
            time.sleep(self.loop_time)
            aggregator.run()
        aggregator.close()
        return True, "Aggregation has ended"
```
- The new versions of Provider and the G3FileRotator are both "single thread" only.  Because we're in twisted territory, you should say in the docstrings that they are for single-thread use.  This is to help people like me, who are reading your code with thread-safety in mind, to understand what the intention is.
- The docstrings for your classes are not formatted properly... you need a blank line to create a new paragraph.  If you want one big paragraph, please "re-fill".
- Aggregator docstring "New data can be passed to it via a thread-safe queue ~~with the function `process_incoming_data`~~."

jlashner:
Implemented Matthew's feedback. Any other suggestions?

jlashner:
Ok cool. I also want to wait until Brian tests it before merging.

BrianJKoopman:
Ah, the review comments I started when the commit was first made are still there...uh, feel free to address if you'd like.

Issue #77, opened 2019-10-20
Site parser


jlashner:
I think it would be useful if instead of having to call `site_config.reparse_args`, we created a `SiteParser` class that subclassed `argparse.ArgumentParser`, which calls `parse_args` on the combination of command line arguments and the arguments found in the site-config file. 

This would make it so that any options specified in the parser like `required` or `type` would be applied to site-config args as well. Also you wouldn't have to run `reparse_args`.

jlashner:
Realizing this is the same problem as issue #30. But I think this would be a good solution to that.

mhasself:
Resolved in #82

Issue #30, opened 2019-02-20
Handling of non-string options in site_config


mhasself:
Arguments (options) passed through site_config are not processed by ArgumentParser, and so no type conversion is done on the string arugments.  This means that command line argument "--num-channels 2" can correspond to site_config argument entry ['--num-channels', 2] rather than ['--num-channels', '2'].  This is all confusing and unnatural and unnecessary.  site_config.py should be modified to incorporate site_config arguments through ArgParser functions, so that type conversion (and checking) is done consistently.

mhasself:
Resolved in #82

PR #82, opened 2019-10-29
Adds parse_arguments function to site_config

jlashner:
This parses arguments from both the command line and then site-config
file simultaneously, and passes them through the argparse.ArgumentParser so that they
can be interpreted with the correct argparse options.

Changed fake_data_agent.py to demonstrate how to use it. For instance, in the fake data agent this makes it so num_channels is always interpreted as an `int` regardless of whether it comes from the site_config file or the command line, and enforces that `mode` is actually either `idle` or `acq`.

This should fix issues #77 and #30.

### DIFF
--- a/ocs/site_config.py
+++ b/ocs/site_config.py
@@ -7,6 +7,7 @@ import sys
 import yaml
 import argparse
 import collections
+import deprecation
 
     
 class SiteConfig:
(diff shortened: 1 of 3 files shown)


ocs/site_config.py
```diff
@@ -430,6 +432,32 @@ def get_config(args, agent_class=None):
...
+    if args.site_realm is None:
+        args.site_realm = site.hub.data['wamp_realm']
+    if args.address_root is None:
+        args.address_root = site.hub.data['address_root']
+    if args.registry_address is None:
+        args.registry_address = site.hub.data.get('registry_address')
+    if (args.log_dir is None) and (host is not None):
+        args.log_dir = host.log_dir
+
 def reparse_args(args, agent_class=None):
```
mhasself: Perhaps mark this function, with a comment or in the docstring, as deprecated in favor of the simpler site_config.parse_args call.

ocs/site_config.py
```diff
@@ -575,4 +592,104 @@ def scan_for_agents(do_registration=True):
...
+    pre_args, _ = pre_parser.parse_known_args(args=sys.argv[1:])
+
+    site, host, instance = get_config(pre_args, agent_class=agent_class)
+
+    # Container from command line args
+    cl_container = ArgContainer(sys.argv[1:])
+
+    arg_list = []
+    for arg in instance.arguments:
+        arg_list.extend(map(str, arg))
```
mhasself: An interesting consequence of this (all this) is that the site_config yaml arguments don't have to be organized as key-value pairs.  Our current syntax of
```
       'arguments': [['--initial-state', 'record'],
                     ['--data-dir', './data']]
```
is now equivalent to:
```
       'arguments': [['--initial-state', 'record', '--data-dir', './data']]
```
but the following, more sensible looking thing, will cause trouble:
```
       'arguments': ['--initial-state', 'record', '--data-dir', './data']
```
It might be worth allowing this latter case to work too... you can just check for isinstance(arg, str).

jlashner:
Awesome, thanks for the feedback. Just committed changes and updated documentation.

PR #85, opened 2019-11-01
Aggregator session data

jlashner:
This PR starts to add aggregator data to the session object. Currently, the session.data object contains 
 - The current file being written
 - The last refresh time, last block_name, and is_stale for every provider that it is monitoring.

I figured we want to monitor providers even after they're stale so that their data shows up in the session object, so I changed when providers were removed. In the new version, stale providers are not removed, and providers are only removed when the aggregator detects a new provider of the same address with a larger (newer) session_id. This way the aggregator will keep track of all providers it has detected throughout its lifetime, but if a new provider with the same address is detected, the old one will be removed and replaced. I think this is ok, because its not possible to have multiple agents with the same instance-id running simultaneously.

@mhasself If there's more info you want me to include in the session data variable for the web monitor let me know!

### DIFF
--- a/agents/aggregator/aggregator_agent.py
+++ b/agents/aggregator/aggregator_agent.py
@@ -76,7 +76,12 @@ class AggregatorAgent:
         session.set_status('starting')
         self.aggregate = True
 
-        aggregator = Aggregator(self.incoming_data, self.time_per_file, self.data_dir)
+        aggregator = Aggregator(
+            self.incoming_data,
+            self.time_per_file,
+            self.data_dir,
+            session=session
+        )
 
         session.set_status('running')
         while self.aggregate:
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -93,6 +93,9 @@ class Provider:
         last_refresh (time):
             Time when the provider was last refreshed (either through data or
             agent heartbeat).
+        last_block_received (str):
+            String of the last block_name received.
+
         log (txaio.Logger):
             txaio logger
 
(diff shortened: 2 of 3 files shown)


agents/aggregator/aggregator_agent.py
```diff
@@ -340,16 +349,37 @@ def process_incoming_data(self):
...
+                return
+
+            self.providers[pid].write(data)
 
     def add_provider(self, prov_address, prov_sessid, frame_length):
         """
         Registers a new provider and writes a status frame.
+        If a provider with prov_address already exists, and the prov_sessid
+        is greater than the current provider's session_id, the old provider
+        will be written to disk and replaced with
```
BrianJKoopman: Incomplete sentence.

mhasself:
I think we do need to retire providers, at least in the G3 frame-stream, rather than letting their prov_id linger forever.  This is for the sake of the output frame-stream, which should be informative, to its consumers, that they can stop tracking a given provider.

I think you're right that the status info in .data is more useful it contains some kind of list of expired providers -- rather than just dropping them and forgetting about them.  But that should not cause a change in how the frames are written.

Perhaps now is a good time to start accepting configuration, in each feed, about the time limit after which a feed should be considered stale.  This would have to be an optional setting for now, with some reasonable default (1 minute or 3 minutes ... whatever makes most sense given the agents currently in distribution).

jlashner:
Just pushed some commits that address the concerns Matthew had with this PR before. Namely, instead of changing the way providers are retired, I just keep a `provider_archive` dict that saves the Provider for the most recent session recorded for each agent address, and keeps even after that provider is retired.

Also I make it so the `fresh_time` is settable at feed registration through the `agg_params` variable, so we can change it on a per agent basis.

jlashner:
Great thanks! Filled in missing docs.

PR #91, opened 2019-11-06
www/monitor.html: fix close/open and view session data

### LINKED ISSUE #87: www monitor uses 100% CPU until connection is made
I just noticed that idling on the www monitor.html page will cause a firefox process to use 100% CPU until a connection to the crossbar server is made. Maybe there's a rapid connection attempt?

mhasself:
I believe this should fix #87 .

Also it anticipates having interesting .data in Operation sessions, as proposed in #85 .

### DIFF
--- a/www/js/ocsbow.js
+++ b/www/js/ocsbow.js
@@ -51,7 +51,7 @@ function AgentClient(_ocs, address) {
     this.feeds = null;
     this.messages = null;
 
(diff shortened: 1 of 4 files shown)


mhasself:
Ya... the issue was that the code could not tell the difference between a realm connection failure and a requested reconnection (which is needed to change the subscribed feed).

PR #95, opened 2019-11-25
InfluxDB Publisher

BrianJKoopman:
This PR introduces the InfluxDB Publisher Agent, which publishes recorded OCS feeds to an instances of InfluxDB. This allows users to use InfluxDB as a backend for Grafana.


BrianJKoopman:
Self-merging to build first official image, comments still certainly welcome.

Issue #72, opened 2019-09-03
HK Aggregator is sensitive to data types passed to it


BrianJKoopman:
If an Agent improperly prepares a data message to pass to the OCS Feed, for instance if a string is passed in a data array, then the aggregator crashes. 

Here's an excerpt from the aggregator logs when a string is passed within a data block (as the bluefors agent was doing until fixed, PR incoming):
```
2019-09-03T21:50:42+0000 Added provider observatory.bluefors.feeds.bluefors (session_id: 1567538827.1608953) with id 6
2019-09-03T21:50:43+0000 Creating file: /data/15675/2019-09-03-21-50-43.g3
2019-09-03T21:51:32+0000 record:1 Crash in thread: [Failure instance: Traceback: <class 'TypeError'>: Invalid assignment
/usr/lib/python3.6/threading.py:916:_bootstrap_inner
/usr/lib/python3.6/threading.py:864:run
/usr/local/lib/python3.6/dist-packages/twisted/_threads/_threadworker.py:46:work
/usr/local/lib/python3.6/dist-packages/twisted/_threads/_team.py:190:doWork
--- <exception caught here> ---
/usr/local/lib/python3.6/dist-packages/twisted/python/threadpool.py:250:inContext
/usr/local/lib/python3.6/dist-packages/twisted/python/threadpool.py:266:<lambda>
/usr/local/lib/python3.6/dist-packages/twisted/python/context.py:122:callWithContext
/usr/local/lib/python3.6/dist-packages/twisted/python/context.py:85:callWithContext
aggregator_agent.py:436:start_aggregate
aggregator_agent.py:111:to_frame
]
2019-09-03T21:51:32+0000 record:1 Status is now "done".
```

A few things could fix this. Type checking in the aggregator before this assignment. Also, stricter specifications on what can go into a data block and perhaps a check in the OCS Feed class. This should be pretty high priority to fix.

BrianJKoopman:
Should be fixed by a combination of #103 and #106.

PR #103, opened 2020-01-31
Aggregator type checking

BrianJKoopman:
This is meant to address #72, or at least start to. A lot of this PR is some code organization and setup for testing the aggregator code. Since this depends on so3g we move to a similar testing scheme as used in the so3g repo where we run tests within a Docker container, then do our reporting outside the container.

The actual fix and related testing is in e686a8a, and will catch the one particular situation we were running into with the Bluefors agent (and perhaps what might end up being the most common mistake in Agents passing data to an OCS Feed), which was that a str instead of a float was being passed. IrregBlockDouble can't store this and so we'd crash.

Currently, in the event we can't cast to float we log the error and move on. This should allow the aggregator to keep on aggregating, even if one or more Agents sends malformed data, however this malformed data will be dropped. I'm open to suggestions for how best to handle this (for instance, when I first wrote these tests I thought it might be better to raise a new TypeError, but that's also going to cause a crash...)

Documentation should still be added to clarify to OCS Agent writers the types of data that can be passed to a Feed. I think currently the best place for that is in the socs guide for writing an Agent though, so I will do so there.


ocs/agent/aggregator.py
```diff
@@ -0,0 +1,471 @@
...
+        for block_name, block in self.blocks.items():
+            if not block.timestamps:
+                continue
+
+            hk = so3g.IrregBlockDouble()
+            hk.prefix = block.prefix
+            hk.t = block.timestamps
+            for key, ts in block.data.items():
+                try:
+                    hk.data[key] = ts
```
jlashner: I realize that this is raising a type error because hk.data is not a regular python dictionary, but when reading the code this is kind of confusing. Can we explicitly check the type instead of using a try except? Also this doesn't need to be this PR, but we should switch this to use G3Ndarry pretty soon

ocs/agent/aggregator.py
```diff
@@ -0,0 +1,471 @@
...
+                except TypeError:
+                    try:
+                        hk.data[key] = [float(x) for x in ts]
```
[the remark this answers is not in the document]
BrianJKoopman: Yeah, point taken. I think it would be clearer to the user if this error was somehow raised by the Agent passing the value. Perhaps additional type checking can be performed in the Feeds code.

Anyway, I removed the cast attempt.

ocs/agent/aggregator.py
BrianJKoopman: This I disagree on. The statement in the try block should work all the time with the exception of the given error (a non-float value was passed within a list to the aggregator.) The alternative is to iteratively check all types in the list passed to the aggregator for all values passed.

```
for key, ts in block.data.items():
    try:
        hk.data[key] = ts
    except TypeError:
        self.log.error("datapoint passed from address " +
                       "{a} to the Provider feed is of " +
                       "invalid type {t}",
                       a=self.address, t=type(ts))
```
vs.

```
for key, ts in block.data.items():
    if all(isinstance(x, float) for x in ts):
        hk.data[key] = ts
    else:
        elf.log.error("datapoint passed from address " +
                        "{a} to the Provider feed is of " +
                        "invalid type {t}",
                        a=self.address, t=type(ts))
```

This did make me realize that my error ultimately wasn't very descriptive though, as the important types are within the list passed, not the type of `ts` itself (which is going to be a list.) I've updated that in 9038495.

BrianJKoopman:
> I haven't actually run the tests, but this looks pretty good and I'm super glad that we can now actually test the aggregator. I have a few comments on how the type checking works that I commented below. I'm not really sure if I agree with attempting to convert strings to floats if the type checking fails. I'd rather just throw an error and not write anything and have them fix the agent than have people not know that they're sending strings instead of floats. Everything else looks good though

Thanks for the feedback. Code updated and comments responded to.

The great part about this is that you don't have to run the tests yourself, Travis ran them for you: https://travis-ci.com/simonsobs/ocs/builds/148282379

I do think we should try to raise this error on the Agent end of things, that will be more obvious to a user developing an Agent than an error logged in the aggregator logs. It might not immediately occur for a user to look there when they aren't seeing data show up.

Dangerously, data might be showing up in Grafana/InfluxDB, because that backend will happily accept a string and show you the value. This might lead to the aggregator raising these errors going unnoticed until one goes to look at the data. That's freaking me out enough that I'm going to open another issue related to enforcing this in at the Feed level right now.

BrianJKoopman:
Merging now before also merging #106. Thanks for the feedback!

Issue #81, opened 2019-10-28
Add documentation for MatchedClient


jlashner:
There should be some documentation for how to use MatchedClient...

PR #104, opened 2020-02-10
Documentation Overhaul

BrianJKoopman:
This is a pretty large overhaul of the OCS documentation. 

### Summary
* Add top level captions to separate "User Guide", "Agent Reference", "Developer Guide" and "API" in the sidebar via toctree captions.
* Re-write most of the introduction to give users a quick introduction to some terms and to mention Docker and its relation to OCS.
* Re-write installation page, pulling a lot from the live monitor setup guide in sisock. This allows OCS docs to stand alone with setup of Docker and software/hardware requirements. 
* Re-work quick start (which referenced old Makefile driven example). Now includes entirely self-contained Docker deployment of an example OCS setup complete with OCS Web, InfluxDB, and Grafana.
* Split "developer" and "user" info about OCS site configs to two pages.
* Update "user" ocs site configs page with more detail about the configuration, and replaced "river_bank" example name with concrete examples.
* Add docker configuration page to describe how to configure OCS Docker containers.
* Split Host Master Agent to the Agent Reference section.
* Split OCS Feeds section to its own page.
* Write a developer guide for writing clients, both in the old style and in the MatchedCilent style.

### Questions
* We're pointing to socs docs for network configuration details, should we move that to the OCS docs too? This is probably generally interesting to OCS users, not just SO.
* Similarly, I would like to move the socs page for ["Creating New Agents"](https://socs.readthedocs.io/en/latest/new_agents.html) to the developer guide here. As that is useful for any OCS user trying to write a new Agent. Any objections?
* Does this satisfy #81? If not, what would you like to see added?


docs/user/site_config.rst
```diff
@@ -0,0 +1,226 @@
...
+telescope.yaml --site-host host-1`` *. One exception to this is when using*
+``--site=none``. *)*
+
+1. Because there are two instances of "Lakeshore240Agent" registered
+   in the SCF, we must somehow pick one when running the agent::
+
+     $ python LS240_agent.py --instance-id=thermo1
+     I am in charge of device with serial number: LSA11AA
+
+2. We can ask our agent to connect to a different WAMP realm, for
```
jlashner: Out of curiosity is there any case which someone would actually do this?

docs/user/docker_config.rst
```diff
@@ -0,0 +1,158 @@
...
+    depends_on:
+      - "sisock-crossbar"
+      - "database"
+
+The top line, ``example-container-name``, defines the name of the service to
+docker-compose. These must be unique. ``image`` defines the docker image used
+for the container. A container can be thought of as a copy of an image. The
+container is what actually runs when you startup your docker service.
+``restart`` allows you to define when a container can be automatically
+restarted, in this instance, always. ``hostname`` defines the hostname internal
```
jlashner: Does this mean whenever `docker-compose up` is run the container is restarted?

docs/user/docker_config.rst
```diff
@@ -0,0 +1,158 @@
...
+defines the version of the docker-compose file format, which corresponds to the
+Docker Engine version you are running. You likely do not have to change this.
+
+Every block below ``services:`` defines a Docker container. Let's look at one
+example container configuration. This example does not represent something we
+would want to actually use, but contains configuration lines relevant to many
+other container configurations::
+
+  example-container-name:
+    image: simonsobs/example-docker-image:latest
```
jlashner: Maybe we should have a note or a warning about using the latest tag instead of an actual version number

docs/user/docker_config.rst
```diff
@@ -0,0 +1,158 @@
...
+container. This is used for configuring behaviors inside the containers.
+``depends_on`` means Docker Compose will wait for the listed containers to
+start before starting this container. This does not mean the services will be
+ready, but the container will be started.
+
+.. note::
+    Environment variables can be used within a docker-compose configuration
+    file. This is done for the `OCS_CONFIG_DIR` mount for the OCS agents in the
+    default template.  For more information see the `docker compose
+    documentation`_.
```
jlashner: Could mention here that you need to run `sudo -E` if you are not part of the docker group.

docs/user/installation.rst
```diff
@@ -0,0 +1,147 @@
...
+    manuiplates iptables to open ports and so a software firewall is likely not
+    enough.
+
+.. note::
+    If you do not have a public IP, but do have access to a gateway to
+    your private network, then port forwarding can be used to view the live monitor
+    remotely.
+
+For examples of possible network configurations see the socs documentation -
+`Network Configuration`_.
```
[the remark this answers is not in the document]
BrianJKoopman: Really? It works in my builds. The corresponding link is at the bottom of the page (and I'm going to move all the other ones there too.)

docs/user/docker_config.rst
BrianJKoopman: Great suggestion! I'm always forgetting to throw that flag, even in documentation apparently! Added.

docs/user/docker_config.rst
BrianJKoopman: No, a `docker-compose up` will only restart a container if a new image has been pulled for it, or if a container that the one in question "depends on" is restarted.

`restart: always` will make sure the container is started on boot is the major thing for me. Other options are 'on-failure' and 'unless-stopped'. Perhaps 'always' is a bit aggressive, but I haven't run into any instances where it's been a terrible thing.

https://docs.docker.com/compose/compose-file/compose-file-v2/#restart

BrianJKoopman: Note, this has been replaced in more recent version of the compose standard with `restart_policy` which has some more control over when they're restarted: https://docs.docker.com/compose/compose-file/#restart_policy

docs/user/site_config.rst
BrianJKoopman: Probably not. This section was straight from the old docs on site configuration, I just made the examples a little more concrete (`LS240_agent.py` vs `river_agent.py`.)

Do you think it should be removed?

docs/user/quickstart.rst
```diff
@@ -0,0 +1,259 @@
...
+    A crossbar configuration cannot be generated without this info.
+    [koopman@vesper self-contained-quickstart]$ ocsbow crossbar generate_config
+    The crossbar config-dir is set to:
+      ./dot_crossbar/
+    Using
+      ./dot_crossbar/config.json
+    as the target output file.
+    
+    Generating crossbar config text.
+    Wrote ./dot_crossbar/config.json
```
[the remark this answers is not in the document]
BrianJKoopman: Good catch, didn't mean to copy the failure part. (The failure was caused by the `OCS_CONFIG_DIR` still pointing to my typical config, not this one for the example.) I removed the failed attempt in the snippet.

docs/user/quickstart.rst
```diff
@@ -0,0 +1,259 @@
...
+                         ['--port', 'tcp://ls240-sim:1094'],
+                         ['--mode', 'acq']]},
+        ]   
+      }
+
+Next, we need to define the Docker Compose file. Again, use this file unchanged.
+
+**docker-compose.yaml**::
+
+    version: '2'
```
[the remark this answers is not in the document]
BrianJKoopman: My approach with compose files has been to keep whatever version they're set to until I need a feature implemented in a new version. Old versions seem to remain compatible. Have you run into problems here?

docs/user/docker_config.rst
BrianJKoopman: Great feedback -- added a warning about docker tags with a brief description of how to find them.

docs/user/quickstart.rst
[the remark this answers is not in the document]
jlashner: No real problems, just like you said sometimes there's a feature that I want to use that doesn't work with the older version, and I can't see any reason to suggest the older one over the newer one. Also later on in the docs an example docker-compose uses 3.7, but its not a huge deal.

docs/user/installation.rst
```diff
@@ -0,0 +1,147 @@
...
+.. _twisted: https://twistedmatrix.com/trac/
+.. _wampy: https://github.com/noisyboiler/wampy
+
+All dependencies are automatically installed via the ``requirements.txt`` file.
+
+Hardware Requirements
+---------------------
+
+You will need a Linux computer running Ubuntu 18.04. Other
+Operating Systems can be used, but are not officially supported.
```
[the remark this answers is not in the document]
BrianJKoopman: Right, good point. I do agree this should be re-worked. This was almost straight from the sisock live monitor guide which was a bit more focused. I'll work on that.

BrianJKoopman:
Thanks for the great feedback Jack! Any thoughts on the questions posed at the end of my original post? Particularly the last one about the MatchedClient docs.

jlashner:
Yeah, I think it satisfies that issue for the time being. There are a few things that I want to address, but possibly not in the PR. I haven't experimented with it much, but MatchedClient can theoretically be run over wampy instead of http if you pass it `client_type='wampy'`, so I think you can also use it for all of your async needs in the standard twisted-client paradigm. But we don't need to edit the documentation on that until I actually test it out.

I like the idea of moving the network configuration instructions to the OCS repo because I think it belongs with all of the other setup instructions, and does not seem specific to SO. Same with the "Create new agent" page, since that describes a lot of core ocs functionality which isn't really described anywhere else. 

Also the link you posted for the Create New Agents page isn't working for me either. It's sending me to "https://socs.readthedocs.io/en/latest/new_agents.html" instead of "https://socs.readthedocs.io/en/latest/user/new_agents.html" which currently works for me. Not sure why ours are different, but it could be the same thing happening with the link in the docs.

docs/user/site_config.rst
jlashner: I think it can be left in, I was just wondering.

BrianJKoopman:
> Yeah, I think it satisfies that issue for the time being.

Great, I'll link the issue then so it closes once this is merged.

> I like the idea of moving the network configuration instructions to the OCS repo because I think it belongs with all of the other setup instructions, and does not seem specific to SO. Same with the "Create new agent" page, since that describes a lot of core ocs functionality which isn't really described anywhere else.

Alright, I'll that in a separate PR/commit.

> Also the link you posted for the Create New Agents page isn't working for me either. It's sending me to "https://socs.readthedocs.io/en/latest/new_agents.html" instead of "https://socs.readthedocs.io/en/latest/user/new_agents.html" which currently works for me. Not sure why ours are different, but it could be the same thing happening with the link in the docs.

Ah, yeah, I tried to clean up the organization in socs' docs too so that the sidebar was also split into sections, which broke that link. You're right in that this also breaks the Network Configuration link, but that happened after the first time it didn't work for you. I just pushed an update, 95f3380, that should fix the link.

docs/user/installation.rst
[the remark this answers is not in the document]
BrianJKoopman: Alright, updated in 4f77609. I split the dependencies page off from installation, as the page was getting quite long, and I had previously split it up in the live monitor guide.

docs/user/quickstart.rst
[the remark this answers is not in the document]
BrianJKoopman: Hmm, I'm a bit unsure of how to address this, as I've mostly run with whatever version I've copied from an example, which as mostly been '2'. '3.7' does seem to work even with syntax from '2'.

I've rolled back my stray '3.7' update in the example you referenced, as I know '2' works in all the Yale configs similar to that example. Updating to recent versions might be good, but needs testing and can perhaps be done outside this PR.

PR #106, opened 2020-02-12
Feed type checking

### LINKED ISSUE #105: Check types of data being passed to OCS Feeds
As brought up in #103, I think we should raise an error to the user at the Agent when an improper type is passed through an OCS feed at `publish_to_feed` if possible. This should ensure uniformity for compatible data types on the backend.

BrianJKoopman:
This is related to #103 and fixes #105.

After we identified that strings were being accidentally passed from the Bluefors log Agent, and that doing so would cause the Aggregator to crash, we've been in need of a mechanism for verifying the data we're passing to an OCS Feed.

This commit PR is based on the aggregator-type-checking branch, since the automated testing depends on the work I did there, so likely that PR should be merged first. But either way, this adds the `ocs.ocs_feed.Feed.verify_message_data_type()` function, which checks the types of data contained in a message's 'data' dictionary.

We also add tests for known valid and invalid cases.

I'd like feedback on the `valid_types` list, is `(int, float)` inclusive enough?

### DIFF
--- a/ocs/ocs_feed.py
+++ b/ocs/ocs_feed.py
@@ -237,5 +237,11 @@ class Feed:
         valid_types = (float, int)
 
         for k, v in message['data'].items():
-            if not all(isinstance(x, valid_types) for x in v):
-                raise ValueError("message 'data' block contains invalid data type")
+            # multi-sample check
+            if isinstance(v, list):
+                if not all(isinstance(x, valid_types) for x in v):
+                    raise ValueError("message 'data' block contains invalid data type")
+            # single sample check
+            else:
+                if not isinstance(v, valid_types):
+                    raise ValueError("message 'data' block contains invalid data type")
--- a/ocs/ocs_feed.py
+++ b/ocs/ocs_feed.py
@@ -193,6 +193,9 @@ class Feed:
                                           timestamp=timestamp)
 
         if self.record:
+            # check message contents
+            Feed.verify_message_data_type(message)
+
             # Data is stored in Block objects
             block_name = message['block_name']
             try:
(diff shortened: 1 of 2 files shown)


jlashner:
Right now that's all that is supported, but I thought we would eventually be aggregating log statements and strings. Does the G3NDarray currently support strings? If so I think we should think ahead a bit and support that at the feed level. Maybe on registering the feed it can set what types are allowed?

ocs/ocs_feed.py
```diff
@@ -219,3 +222,26 @@ def publish_message(self, message, timestamp=None):
...
+                Data to be published (see Feed.publish_message for details).
+
+        """
+        valid_types = (float, int)
+
+        for k, v in message['data'].items():
+            # multi-sample check
+            if isinstance(v, list):
+                if not all(isinstance(x, valid_types) for x in v):
+                    raise ValueError("message 'data' block contains invalid data type")
```
jlashner: I think this error message should be a bit more informative. Can we log the feed_name, block_name, data field, value, and type that's being passed? Or do you think that's unnecessary info?

BrianJKoopman:
> Right now that's all that is supported, but I thought we would eventually be aggregating log statements and strings. Does the G3NDarray currently support strings? If so I think we should think ahead a bit and support that at the feed level. Maybe on registering the feed it can set what types are allowed, and have it default to (float, int)?

I'm not sure if G3NDarray supports strings, I'll look into it.

My motivation for this has been to protect the g3 Aggregator from data that's going to cause it to crash. Allowing the user (whoever is writing the Agent) to specify the allowed types puts us back in a similar situation to now, where the user needs to know what types the Aggregator supports. Right now a user is free to pass strings into the Feed. Given the option to add `str` to the list of supported types on their own they could "solve" an unsupported TypeError by allowing strings in the Agent. So I'm less inclined to allow it for now until some future PR that updates the Aggregator to accommodate storing strings/logs.

On a related note though, documenting supported Aggregator types in a clear way should also be a priority.

jlashner:
awesome, I'm cool with merging this now and then coming back to it when/if we add string support in the aggregator

ocs/ocs_feed.py
BrianJKoopman: I did add the invalid types passed to the error. I debated on the other info. Sort of working off of examples in the Python standard library for TypeErrors. I think it's reasonable to expect some level of debugging from a user writing an Agent that hits this TypeError. I was also thinking of adding some debug level logging with the other requested info.

BrianJKoopman:
> awesome, I'm cool with merging this now and then coming back to it when/if we add string support in the aggregator

Great. Yeah, the thing to think about then is how to pipe different data to different aggregator back ends that might be running. For instance, in the now common situation where the g3 Aggregator and InfluxDB are running, strings representing numbers are alright for InfluxDB (provided the type stays consistent), but I'm sure logs won't be.

Anyway, now that #103 is merged, I've rebased this branch on master and will now merge it as well. Thanks again for all the feedback!

PR #107, opened 2020-02-17
Updates to g32influx script

ahincks:
As I work on assessing the scalability of `influxdb` as an "official" storage format, I've made some modifications to this script.

Speed is greatly improved (on my laptop, ~5 seconds per file instead of ~1 minute):
- By using line format rather than JSON for writing to `influxdb` DB.
- By writing whole timelines at once time rather than field-by-field.

Rather than using a standalone `sqlite` DB to record which files have been written, there is now a measurement in the `influxdb` DB, `.g32influx.log`, that records when files have been completed (and also records if they have been started, as well as if an error has occurred. This can be queried to determine whether a file needs to be written or not (which can be overridden with `-f`).

In addition to the `feed` tag, an `md5sum` tag is added to each HK point so that the provenance of data can be traced.
 - One thing to test for is whether this increases the series cardinality too much. Based on [this guideline](https://docs.influxdata.com/influxdb/v1.7/guides/hardware_sizing/#general-hardware-guidelines-for-a-single-node) I think it's going to be OK, but will want to assess this.

Although I'm creating this pull request here so that you can see the changes, it could reflect a simple branch of this code away from `ocs` as I continue my own investigation. So, if you have concerns about anything I've done here and don't think this belongs in `ocs`, no problem.

### DIFF
--- a/bin/g32influx
+++ b/bin/g32influx
@@ -186,9 +186,16 @@ class DataLoader:
         filename : str
             Full path to file to publish.
 
+        Returns
+        -------
+        int
+            Return value from scanner.run()
+
         """
         scanner = SingleFileScanner(filename, self.influxclient)
-        scanner.run()
+        rval = scanner.run()
+
+        return rval
 
     def publish_all_files_to_influxdb(self):
         """Publish all files found in target to InfluxDB.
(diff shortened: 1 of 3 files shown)


BrianJKoopman:
Thanks for the PR, Adam! Getting that kind of speed improvement would be great. I'm looking forward to reviewing this, but probably won't be able to look much at it until early next week, after our SMuRF/DAQ meeting at Princeton.

BrianJKoopman:
After talking to @ahincks and incorporating some of the changes here into #153 we agreed to close this PR.

PR #108, opened 2020-02-25
ocs-util docker

jlashner:
This adds an easy ocs-util docker, which can be used to run general ocs/so3g jupyter notebooks, and be used with 'docker exec' to run other commands such as spt3g-dump (this might not actually work currently since it requires switching the spt3g-dump file to use python3).

If you add this to your docker-compose it'll automatically start a jupyter notebook server in the "/data" directory without any additional setup needed:
```
  ocs-util:
    image: ocs-util:latest
    environment: 
      JUPYTER_PORT: 8880
      JUPYTER_PW: ocs_password
    volumes:
      - ${OCS_CONFIG_DIR}:/config
      - /data:/data
    ports:
      - 8880:8880
```
It will run on port 8880, and the password will be ocs_password. If you create a home directory for the ocs user and mount that in also, it'll save the password cookie so you don't have to enter it every time you restart it.

I think this would be really nice for people who don't have ocs/so3g installed on their systems so they can still analyze data. Let me know what you think.


mhasself:
I'm a fan of this, but can you put the documentation (such as the docker-compose block you have in the PR text...) into the docs/ proper?

BrianJKoopman:
Yeah, agreed the docs should be updated.

Is this meant to replace the so3g-jupyter container?

jlashner:
Ok added documentation. Also I wanted vim in the ocs-util docker, but thought it would be useful in all of the dockers since its something I always try to use and get frustrated that it doesn't exist, so I added it to the base ocs docker.

jlashner:
Added changes we talked about last week.

PR #110, opened 2020-02-25
Create script to convert .g3 filenames to new ctime format

BrianJKoopman:
At the SMuRF/DAQ Workshop we talked about renaming old .g3 files from the `"%Y-%m-%d-%H-%M-%S"` format to one based on ctime `ctime.g3`.

This PR introduces `datestring2ctime`, a script in `ocs/bin/` that converts all .g3 files in a given directory (also works on single files if you want to do that for some reason).

We also change the naming format in the Aggregator itself. Note, a similar change needs to happen on the socs timestream-aggregator.

### DIFF
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -51,9 +51,6 @@ def make_filename(base_dir, make_subdirs=True):
             True if func should automatically create non-existing subdirs.
     """
     start_time = time.time()
-    start_datetime = datetime.datetime.fromtimestamp(
-        start_time, tz=datetime.timezone.utc
-    )
 
     subdir = os.path.join(base_dir, "{:.5}".format(str(start_time)))
 
@@ -64,7 +61,7 @@ def make_filename(base_dir, make_subdirs=True):
             raise FileNotFoundError("Subdir {} does not exist"
                                     .format(subdir))
 
-    time_string = int(start_datetime.timestamp())
+    time_string = int(start_time)
     filename = os.path.join(subdir, "{}.g3".format(time_string))
     return filename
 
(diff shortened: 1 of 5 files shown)


ocs/agent/aggregator.py
```diff
@@ -64,7 +64,7 @@ def make_filename(base_dir, make_subdirs=True):
             raise FileNotFoundError("Subdir {} does not exist"
                                     .format(subdir))
 
-    time_string = start_datetime.strftime("%Y-%m-%d-%H-%M-%S")
+    time_string = int(start_datetime.timestamp())
```
mhasself: You don't need the start_datetime variable any more.

ocs/rename.py
```diff
@@ -0,0 +1,154 @@
...
+        try:
+            basename = os.path.basename(f)
+            datetime.datetime.strptime(basename, "%Y-%m-%d-%H-%M-%S.g3")
+            new_list.append(f)
+        except ValueError:
+            if verbose is not None:
+                print(f"{f} does not match datestring format, removing from list.")
+            else:
+                pass
+
```
mhasself: "else: pass" is the same as ""

ocs/agent/aggregator.py
BrianJKoopman: Right, good point. Removed.

ocs/rename.py
BrianJKoopman: Right, good point. Removed.

BrianJKoopman:
Yeah, I ran it on the entire Yale data set, which goes all the way back to 2019-03-25. I will make that recommendation, as I did the same with my own data.

PR #111, opened 2020-02-28
Clean up some old files in the aggregator directory

BrianJKoopman:
I've been meaning to do this every time I see those .DS_Store files. The one possibly controversial deletion is the .ipynb, but I think I was the one that asked for the example for how to open a .g3 file back then.


jlashner:
Yep this seems ok to me. I'd say go ahead and merge.

PR #114, opened 2020-03-09
Cleanup broken functions and scripts

jlashner:
This PR is to remove functions and scripts in the repository that no longer work, and finally closing issue #9...

This consists mainly of the following:
 - various `ocs.cfg` files in the aggregator directory and in the clients directory
 - `ocs_agent.init_ocs_agent` and `client_t.run_control_script`
 - lots of clients that use an old `aggregator` interface
 - the func `get_ocs_config` from `ocs/base.py`

The only other agent using the old `init_ocs_agent` function is in the `socs/agents/smurf` directory which is way out of date anyway... I'll submit a PR in socs removing that.

This also renames the `run_control_script2` function to `run_control_script`. Are the remaining clients in this repo still being used? Can they be moved somewhere else since the lakeshore clients aren't even in ocs?

Finally removing all of this stuff is so cathartic... Let me know if you think anything I removed is problematic and should be put back in.


jlashner:
Should I just remove the clients directory?

BrianJKoopman:
> Should I just remove the clients directory?

Fine by me.

PR #116, opened 2020-03-31
MatchedClient: wrap replies with a nice __repr__ for interactive use

mhasself:
MatchedClient operation calls return an object with a nice ```__repr__``` string, e.g.:

```
  >>> c.set_autoscan.wait()
  OCSReply: OK : Operation "set_autoscan" just exited.
    set_autoscan[session=7]; status=done without error 30.6 s ago, took 0.113400 s
    messages (4 of 4):
      1585667844.423 Status is now "starting".
      1585667844.424 Status is now "running".
      1585667844.535 Set autoscan to True
      1585667844.536 Status is now "done".
    other keys in .status: data
```

This still works with the form "ok, msg, status = op.status()", for
example, because the OCSReply object is still a tuple, deep down...

### DIFF
--- a/docs/developer/clients.rst
+++ b/docs/developer/clients.rst
@@ -174,7 +174,7 @@ parameter expansion syntax::
 Replies from Operation methods
 ==============================
 
-The responses from Operation methods is a tuple, (success, message,
+The responses from Operation methods is a tuple, (status, message,
 session).  The elements of the tuple are:
 
   ``status``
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
This is great!

It'd also be great to add an example of this output and a description of how to check the status of your task to the MatchedClient documentation, that's currently missing.

docs/developer/clients.rst
```diff
@@ -168,3 +168,79 @@ parameter expansion syntax::
     therm_client.init.start(**params)
     therm_client.init.wait()
     time.sleep(.05)
+
+
+
+Replies from Operation methods
+==============================
+
+The responses from Operation methods is a tuple, (success, message,
```
BrianJKoopman, line 177: `s/success/status`?

PR #117, opened 2020-03-31
MatchedClient: make MatchedTask and MatchedProcess callable

mhasself:
This is another suggestion for improving interactive use / scripting.  Calling start() and wait() separately when you want to run a Task is an important option to have, in our asynchronous architecture.  But in practice it is often just annoying.

For a Process, start() and wait() is not as useful a shortcut.  So instead we just have ```__call__`` map to status().

I acknowledge that having ```__call__``` do different things for Tasks and Processes is slightly weird.  But Tasks and Processes are different!

### DIFF
--- a/docs/developer/clients.rst
+++ b/docs/developer/clients.rst
@@ -1,4 +1,4 @@
-.. highlight:: rst
+.. highlight:: python
 
 .. _clients:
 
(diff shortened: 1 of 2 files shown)


mhasself:
@BrianJKoopman  Indeed... docs are in now (both PRs).

I can handle the merges... there will likely be some conflicts.

Issue #115, opened 2020-03-30
TimeoutLock Documentation


jlashner:
ocs_twisted.TimeoutLock desperately needs documentation.

BrianJKoopman:
We should autodoc the class in the read the docs before closing.

PR #125, opened 2020-04-16
Merges yielding-lock into timeout-lock

### LINKED ISSUE #121: Improved locks for general use
In https://github.com/simonsobs/socs/pull/69 we defined a replacement for TimeoutLock (with a slightly different interface...) that is better suited for situations where a sampler yields the lock occasionally to allow other threads to access hardware.  This should be moved into OCS proper, and perhaps replace TimeoutLock entirely (discuss).

We might even consider a more powerful lock, that can yield preferentially depending on the priority of the request; this is not too difficult to implement but it's not clear the complexity is yet needed anywhere.

jlashner:
This PR is to merge the yielding lock logic from the LS372 agent into the OCS timeout lock and to add documentation. 

Resolves #121 and #115.

### DIFF
--- a/ocs/ocs_twisted.py
+++ b/ocs/ocs_twisted.py
@@ -28,8 +28,9 @@ class TimeoutLock:
                 If set to 0 the acquire calls will be non-blocking and will
                 immediately return the result.
 
-                If set to any value greater than 0, this will block for that
-                many seconds before returning the result.
+                If set to any value greater than 0, this will block until
+                the lock is acquired or the timeout (in seconds) has been
+                reached.
 
                 If set to -1 this call will block indefinitely until the lock
                 has been acquired.
@@ -39,6 +40,8 @@ class TimeoutLock:
 
             job (string, optional):
                 Job name to be associated with current lock acquisition.
+                The current job is stored in self.job so that it can be
+                inspected by other threads.
 
         Returns:
             result (bool): Whether or not lock acquisition was successful.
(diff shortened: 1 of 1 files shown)


ocs/ocs_twisted.py
```diff
@@ -1,29 +1,113 @@
...
+            timeout (float, optional):
+                Sets the timeout for lock acquisition.
+
+                If set to 0 the acquire calls will be non-blocking and will
+                immediately return the result.
 
-    def acquire(self, blocking=True, timeout=-1):
-        return self._lock.acquire(blocking, timeout)
+                If set to any value greater than 0, this will block for that
+                many seconds before returning the result.
```
mhasself, line 32: It will return as soon as it has acquired the lock, or within that many seconds if it is unable to acquire the lock.

ocs/ocs_twisted.py
```diff
@@ -1,29 +1,113 @@
...
+
+                If set to -1 this call will block indefinitely until the lock
+                has been acquired.
+
+                If not set (default), it will use the TimeoutLock's
+                default_value (which itself defaults to 0).
+
+            job (string, optional):
+                Job name to be associated with current lock acquisition.
```
mhasself, line 41: Mention this is stored in self.job for other threads to inspect if need me.

ocs/ocs_twisted.py
```diff
@@ -1,29 +1,113 @@
...
     @contextmanager
-    def acquire_timeout(self, timeout, job='unnamed'):
-        result = self._lock.acquire(timeout=timeout)
+    def acquire_timeout(self, timeout=None, job='unnamed'):
+        """
+        Context manager to acquire and hold a lock.
+
+        Args:
+            timeout (float, optional):
+                Sets the timeout for lock acquisition.
```
mhasself, line 85: I think it's better not to duplicate the full description of timeout, which is the same as for acquire().  Just refer to the acquire method.  More useful here would be to have an example usage -- i.e. showing that within the context you must immediately check the returned value to see if you actually have the lock.

ocs/ocs_twisted.py
```diff
@@ -1,29 +1,113 @@
...
+                Sets the default timeout value for acquire calls.
+                Defaults to 0.
+        """
         self.job = None
-        self._lock = threading.Lock()
+        self._active = threading.Lock()
+        self._next = threading.Lock()
+        self._default_timeout = default_timeout
+
+    def acquire(self, timeout=None, job=None):
```
mhasself, line 20: This is an API change... but acquire_timeout is not.  Is the claim that no one in socs uses "acquire" directly so this won't break anything?

mhasself:
@jlashner Not sure if you're waiting for me on this... but since I approved, but with some minor requests, I think you should feel free to merge once the requests have been addressed (which they seem to have been).

jlashner:
Great, thanks

PR #128, opened 2020-05-07
Switched from IrregBlockDouble to G3Timesample

jlashner:
This PR switches the aggregator to using the G3Timesample object to store data instead of IrregBlockDouble. It uses a helper function `g3_cast` to attempt to cast a python object into the corresponding G3 datatype, with int, str, float --> G3Int, G3String, and G3Double, and lists of a single data type X being casted to G3VectorX. I also add some tests for the `g3_cast` function so you can see cases in which it will/won't work. The type checking also now occurs in this `g3_cast` function, since if the data is of an unsupported type, or is a list of inconsistent types, a TypeError will be raised and handled by the aggregator.

### DIFF
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -1,11 +1,8 @@
 import os
-import datetime
 import binascii
 import time
-
-from typing import Dict
-
 import txaio
+from typing import Dict
 
 from ocs import ocs_feed
 
@@ -18,6 +15,7 @@ else:
     G3Module = object
 
 
+HKAGG_VERSION = 1
 _g3_casts = {
     str: core.G3String, int: core.G3Int, float: core.G3Double,
 }
@@ -305,7 +303,7 @@ class G3FileRotator(G3Module):
         by `filename` function passed to constructor
         """
         for frame in frames:
-
+            frame['hkagg_version'] = HKAGG_VERSION
             ftype = frame['hkagg_type']
 
             if ftype == so3g.HKFrameType.session:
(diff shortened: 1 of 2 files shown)


ocs/agent/aggregator.py
```diff
@@ -18,6 +18,56 @@
...
+        str   -> G3String
+        float -> G3Double
+    and lists of type X will go to G3VectorX. If ``time`` is set to True, will
+    convert to G3Time or G3VectorTime.
+
+    Args:
+        data (int, str, float, or list):
+            Generic data to be converted to a corresponding G3Type.
+        time (bool, optional):
+            If True, will try to cast to G3Time or G3VectorTime.
```
mhasself, line 42: Convert from what to a G3Time?  A unix timestamp?

ocs/agent/aggregator.py
```diff
@@ -18,6 +18,56 @@
...
+                        "be one of {}".format(dtype, _g3_casts.keys()))
+    if is_list:
+        if time:
+            return core.G3VectorTime(list(map(core.G3Time, data)))
+        else:
+            cast = _g3_list_casts[type(data[0])]
+            return cast(data)
+    else:
+        if time:
+            return core.G3Time(data)
```
mhasself, line 66: This won't convert from a unix timestamp to a G3Time... need to scale by G3Units.seconds, or something like that.

jlashner:
Thanks, fixed the G3Time conversion error, but I'm not sure where to change the hkagg version. Do I have to do anything in the aggregator or is that all in the so3g src?

Issue #119, opened 2020-04-02
Set wamp_server and wamp_http for individual hosts


jlashner:
The location of the wamp_server and wamp_http right now differ whether you're running your ocs script from the host (or in a docker with network_mode host) vs within a docker-container. This makes setting these values kind of confusing and makes the arguments to ocs client scripts differ depending on whether they are run in a docker (like ocs-util) vs on the host.

This can be solved by allowing `wamp_server` and `wamp_http` to be set in the default.yaml file on a per-host basis and have that override the one set in the hub section.

BrianJKoopman:
Documented in #276. Going to link this issue to that PR as well.

PR #129, opened 2020-05-12
Allows host config to override wamp_server and wamp_http

jlashner:
A quality of life improvement to allow us to override the wamp_server and wamp_http on a per-host basis (resolving #119). This is needed as the wamp_server and wamp_http addresses differ depending on whether the host is a docker container or not. For instance, crossbar may be accessible through `localhost:8001` if you're on the host, while being accessible through `sisock-crossbar:8001` if within a docker. This means users need to know where an OCS agent or client is being run to know the correct command line arguments to use, and must manually specify the server and http arguments pretty much every time they start an agent or a client.

I had to change `site_config.get_config` a bit, as it was previously not getting the host config dict for control objects, but I've tested my changes and made sure that it still works for both agents and clients. Let me know if there are any issues, thanks!

### DIFF
--- a/ocs/site_config.py
+++ b/ocs/site_config.py
@@ -418,6 +418,36 @@ def get_config(args, agent_class=None):
     # Load the site config file.
     site_config = SiteConfig.from_yaml(site_file)
 
+    # Matching behavior.
+    no_host_match = (agent_class == '*control*')
+    no_dev_match = no_host_match or (agent_class == '*host*')
+
+    # Identify our host and update site.hub.
+    host_config = None
+    if args.site_host is not None:
+        host_attempts = [args.site_host, 'localhost']
+    else:
+        host_attempts = [socket.gethostname(), 'localhost']
+    for host_try in host_attempts:
+        if host_try in site_config.hosts:
+            host_config = site_config.hosts[host_try]
+            host_update_dict = {
+                k: host_config.data[k]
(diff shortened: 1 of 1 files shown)


ocs/site_config.py
```diff
@@ -437,27 +437,29 @@ def get_config(args, agent_class=None):
...
+    if args.working_dir is not None:
+        host_config.working_dir = args.working_dir
+    if args.log_dir is not None:
+        host_config.log_dir = args.log_dir
+    # Overrides wamp_server and wamp_http if specified in host config,
+    # and if command line args are not set
+    if 'wamp_server' in host_config.data and args.site_hub is None:
+        site_config.hub.data['wamp_server'] = host_config.data['wamp_server']
+    if 'wamp_http' in host_config.data and args.site_http is None:
+        site_config.hub.data['wamp_http'] = host_config.data['wamp_http']
```
mhasself, line 462: The 10 lines from here up will fault if host_config is None, which is completely allowed (for control clients for example)... please restore the guard.

jlashner:
Hey Matthew I think I fixed the bug you mentioned. I moved it so all site_config.host updates happen within the `if host_try in site_config.hosts` block, which I think addresses your concern about host_config not being defined for the updates. 

I also moved it to be before the command line args update statements since we want those to take priority. Let me know if there's anything else you think I should change!

PR #131, opened 2020-05-15
Added timeout lock to readthedocs

### LINKED ISSUE #115: TimeoutLock Documentation

jlashner:
Added a TimeoutLock page to the readthedocs. Closes #115.


docs/developer/timeout_lock.rst
```diff
@@ -0,0 +1,52 @@
...
+
+TimeoutLock
+============
+
+Overview
+---------
+
+In OCS, operations are mainly asynchronous, however most agents have
+restrictions on which operations are allowed to be run simultaneously.
+For instance, when only one operation can query with a hardware device at a time
```
BrianJKoopman, line 13: I think this could be reworded, how about:
"For instance, when an Agent interfaces with a hardware device over serial communication, two commands cannot be issued simultaneously."

docs/developer/timeout_lock.rst
```diff
@@ -0,0 +1,52 @@
...
+through serial communication, and so trying to control hardware variables while
+an acquisition process is in progress is dangerous.
+
+In these cases, it is required for agents to create a `lock` which restricts
+what operations can run simultaneously. For instance, in your agent it may
```
BrianJKoopman, line 18: Another wording suggestion for the sentence that starts with "For instance..." -> "In our example, it is necessary for each operation that communicates over with the hardware device to acquire the lock."

docs/developer/timeout_lock.rst
```diff
@@ -0,0 +1,52 @@
...
+be necessary for a process to acquire the lock whenever it talks to the
+hardware. Then when a new process is started, it will either grab the lock
+if no other operation has it, wait a specified amount of time for the lock
+to be free, or fail and return immediately.
+
+The ocs TimeoutLock is a lock that implements a few useful additions to the
+standard ``threading.lock``. Mainly:
+ -  It can be acquired the ``acquire_timeout`` context
```
BrianJKoopman, line 26: There might be a word missing here?

docs/developer/timeout_lock.rst
```diff
@@ -0,0 +1,52 @@
...
+    manager, which will automatically release even if an error is raised.
+ -  A ``job`` string can be set on acquisition, so it is clear why an acquisition
```
BrianJKoopman, line 28: Change after the comma to "describing the job obtaining the lock." This gives developers guidance for what they should call the job.

docs/developer/timeout_lock.rst
```diff
@@ -0,0 +1,52 @@
...
+    failed.
+ -  The ``release_and_acquire`` function can be used to release the lock so that
+    so operations can safely run during long-running processes
+ -  A ``timeout`` can be set to limit how long an operation will wait for the
+    lock before failing.
+
```
BrianJKoopman, line 34: Add new section for Examples (at the '---' level.) Will provide room for when `YieldingLock` gets moved.

docs/developer/timeout_lock.rst
```diff
@@ -0,0 +1,52 @@
...
+            return False
+
+        print("Lock acquired!")
+
+
+API
+----
+
+.. autoclass:: ocs.ocs_twisted.TimeoutLock
+    :members:
```
BrianJKoopman, line 52: The API is autodoc'd already in the [API Reference](https://ocs.readthedocs.io/en/latest/api.html). This causes a bunch of warnings for duplicate autodocs. I've solved this in previous instances by putting `:noindex:` on the page reference, i.e. here. Searches will yield the API page in that case.

docs/index.rst
```diff
@@ -46,6 +46,7 @@ write OCS Agents or Clients.
     developer/feeds
     developer/agents
     developer/clients
+    developer/timeout_lock
```
BrianJKoopman, line 49: I'm not sure this should get its own top level in the sidebar, I think it fits better under Agents. Though as I say that I realize the Agents page should probably be broken up a bit. I do think we should move this in there, whether now or later.

docs/developer/timeout_lock.rst
BrianJKoopman: Ah, in Slack I was told the functionality was moved into TimeoutLock already. A more advanced example of using the lock to allow a long running process that requires the lock to yield to a task that needs to communicate intermittently would be very valuable. Much like is done in the 372 Agent now.

jlashner:
Hey Brian, thanks for the suggestions. I reworded the parts you mentioned. Let me know if you still don't like them. I also put everything into `agents.rst` and added an extra example. While going over the 372's usage I found something that doesn't seem right to me (https://github.com/simonsobs/socs/issues/76), so I changed it a bit in the example.

docs/developer/agents.rst
```diff
@@ -430,3 +430,76 @@ containerized Agents together with the command
 
 Depending on your host's permissions, this command may need to be run with 
 ``sudo``.
+
+.. _timeout_lock
```
BrianJKoopman, line 434: This is missing a colon at the end: It should be:
```
.. _timeout_lock:
```

docs/developer/agents.rst
```diff
@@ -430,3 +430,76 @@ containerized Agents together with the command
...
+For instance, when an Agent interfaces with a hardware device over serial
+communication, two commands cannot be issued simultaneously.
+In these cases, it may be required for agents to create a `lock` which restricts
+what operations can run concurrently.
+Requiring this lock to be acquired before any device communication is one
+way to ensure that serial messages are not crossed, or avoid any other
+dangerous asynchronous behavior.
+
+The ocs :ref:`TimeoutLock <ocs_twisted_api>` is a lock that implements a few useful additions to the
+standard ``threading.lock``. Mainly:
```
BrianJKoopman, line 453: Bullets in .rst aren't indented, this thows an "Unexpected indentation" warning when first compiled. Reformat with a blank line after "Mainly:" and no single space indication on the block.

https://docutils.sourceforge.io/docs/user/rst/quickref.html#bullet-lists

docs/developer/agents.rst
```diff
@@ -430,3 +430,76 @@ containerized Agents together with the command
...
+The following example shows how to acquire the TimeoutLock with the
+``acquire_timeout`` context manager::
+
+    lock = TimeoutLock()
+
+    with lock.acquire_timeout(timeout=3.0, job='acq') as acquired:
+        if not acquired:
+            print(f"Lock could not be acquired because it is held by {lock.job}")
+            return False
+        print("Lock acquired!")
```
BrianJKoopman, line 476: I think a comment with something like "lock requiring actions here" to show the user where to put their code with respect to the lock, i.e. in the with, not after it, would be good.

docs/developer/agents.rst
```diff
@@ -430,3 +430,76 @@ containerized Agents together with the command
...
+use the ``release_and_acquire`` function to allow other operations to interject
+without completely giving up the lock.
+
+This example will acquire a timeout lock before entering an infinite loop of
+"data acquisition". About every second it will release and re-acquire the lock
+so that short tasks can be run without stopping the acquisition. If
+the operation cannot re-acquire the lock, it will print a message and return
+``False``::
+
+    with self.lock.acquire_timeout(timeout=0, job='acq') as acquired:
```
BrianJKoopman, line 491: This example would similarly benefit from where the developer is expect to interject their lock reliant code.

This also references class attributes like `self.lock`, can we adjust the example to establish the two locks before the `with` so that the example stands alone?

jlashner: Was thinking about doing that but using a lock that you initialize right before the `with` statement kind of defeats the purpose of using a lock, so I wouldn't want people to copy and paste that into their code. I could do something like expanding the example so that the lock is initialized globally, and then the acquisition is done within a function.

jlashner:
Alright, did another iteration on the `release_and_acquire` example that shows it's usage in the context of a simple agent with an interrupting operation. Hopefully the example makes more sense within this context.

Issue #130, opened 2020-05-12
Validate field names in hk_aggregator


mhasself:
There are rules for field names defined in the tod_format doc.  We should consider revising them, and then check for validity in the aggregator so invalid fields don't end up in the data.

jlashner:
What rules are you thinking of? I think we should definitely stop using spaces if field names and check against that, but is there anything else in particular?

mhasself:
Current rules are very strict:  _Names of fields and blocks should follow certain rules to be valid. A valid name:_
_– contains only letters (a-z, A-Z; case sensitive), decimal digits (0-9), and the
underscore (\_).
– begins with a letter, or with any number of underscores followed by a letter.
– is no more than 255 characters long._

This is strict... but based on:
- The "." is forbidden because it's a domain separator in OCS agent and feed names.
- One might want to parse arithmetic expressions with field names, so it's perhaps wise to outlaw +/-/*//.
- For easier string parsing of lists of field names, we would be wise to outlaw whitespace and "," and ";".

Ultimately is easier to define by inclusion rather than exclusion, which led to the above rules, which allow for _ as the only separator.

PR #132, opened 2020-05-15
Field name validation

BrianJKoopman:
This PR is to address #130.

We add a field name validation method to the Feed class and use it to raise an error when an Agent publishes data with an invalid field name. This should get the attention of developers if they happen to choose an invalid field name.

Since we were already looping over the contents of the message, we change our type checking method slightly to accommodate checking fields at the same time.

The Aggregator uses the same validation method to check for invalid names. While invalid names shouldn't even make it to the Aggregator now, we check nonetheless. Since we don't want the aggregator to crash or write invalid fields to disk our two options are:
* drop the invalid fields
* try to fix the fields

I opted for fixing the fields, so we remove invalid characters from fields if we see them at the Aggregator, but again, we shouldn't see them. We log an error if we do. This involved rebuilding the data message before building the Block structure when needed.

There are several accompanying tests, mostly for the Feed.

### DIFF
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -150,6 +150,57 @@ class Provider:
 
         return True
 
+    def _verify_provider_data(self, data):
+        """Check the provider data for invalid field names. Meant to be used in
+        combination with Provider._rebuild_invalid_data().
+
+        Args:
+            data (dict): data dictionary passed to Provider.save_to_block()
+
+        Returns:
+            bool: True if all field names valid. False if any invalid names found
+
+        """
+        verified = True
+        for block_name, block_dict in data.items():
+            for field_name, field_values in block_dict['data'].items():
+                try:
+                    ocs_feed.Feed.verify_data_field_string(field_name)
+                except ValueError:
+                    self.log.error("data field name {field} is " +
+                                   "invalid, removing invalid characters.",
+                                   field=field_name)
+                    verified = False
+
+        return verified
+
+    def _rebuild_invalid_data(self, data):
+        """Rebuild an invalid data dictionary.
+
+        Args:
+            data (dict): data dictionary passed to Provider.save_to_block().
+
+        Returns:
+            dict: A rebuilt data dictionary with invalid characters stripped
+                  from the field names, limited to 255 characters in length.
+
+        """
+        new_data = {}
+        for block_name, block_dict in data.items():
+            new_data[block_name] = {}
+            # rebuild block_dict
+            for k, v in block_dict.items():
+                if k == 'data':
+                    new_data[block_name]['data'] = {}
+                    for field_name, field_values in block_dict['data'].items():
+                        # replace invalid characters, and limit to 255 characters
+                        new_field_name = re.sub('[^a-zA-Z0-9_]', '', field_name)[:255]
+                        new_data[block_name]['data'][new_field_name] = field_values
+                else:
+                    new_data[block_name][k] = v
+
+        return new_data
+
     def save_to_block(self, data):
         """Saves a list of data points into blocks. A block will be created
         for any new block_name.
@@ -182,40 +233,13 @@ class Provider:
                 if b['timestamps']:
                     self.frame_start_time = min(self.frame_start_time, b['timestamps'][0])
 
+        self.log.debug('data passed to block: {d}', d=data)
+        verified = self._verify_provider_data(data)
 
-        # We don't want to do this in the original data.items() loop below, but
-        # loop through and see if we need to rebuild the dict, and if we
-        # do,then do it
-        print('before:', data)
-        # fix invalid names...
-        rebuild_data = False
-        for block_name, block_dict in data.items():
-            for field_name, field_values in block_dict['data'].items():
-                try:
-                    ocs_feed.Feed.verify_data_field_string(field_name)
-                except ValueError:
-                    self.log.error("data field name {field} is " +
-                                   "invalid, removing invalid characters.",
-                                   field=field_name)
-                    rebuild_data = True
-
-        new_data = {}
-        if rebuild_data:
-            for block_name, block_dict in data.items():
-                new_data[block_name] = {}
-                # rebuild block_dict
-                for k, v in block_dict.items():
-                    if k == 'data':
-                        new_data[block_name]['data'] = {}
-                        for field_name, field_values in block_dict['data'].items():
-                            new_field_name = re.sub('[^a-zA-Z0-9_]', '', field_name)
-                            new_data[block_name]['data'][new_field_name] = field_values
-                    if k != 'data':
-                        new_data[block_name][k] = v
-
-        print('after:', new_data)
-
-        data = new_data
+        if not verified:
+            self.log.info('rebuilding data containing invalid field name')
+            data = self._rebuild_invalid_data(data)
+            self.log.debug('data after rebuild: {d}', d=data)
 
         for key, block in data.items():
             try:
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -167,7 +167,7 @@ class Provider:
                 try:
                     ocs_feed.Feed.verify_data_field_string(field_name)
                 except ValueError:
-                    self.log.error("data field name {field} is " +
+                    self.log.error("data field name '{field}' is " +
                                    "invalid, removing invalid characters.",
                                    field=field_name)
                     verified = False
@@ -195,8 +195,11 @@ class Provider:
                  through this method, and that should be checked.
 
         """
+        # replace spaces with underscores
+        new_field_name = field_name.replace(' ', '_')
+
         # replace invalid characters
-        new_field_name = re.sub('[^a-zA-Z0-9_]', '', field_name)
+        new_field_name = re.sub('[^a-zA-Z0-9_]', '', new_field_name)
 
         # grab leading underscores
         underscore_search = re.compile('^_*')
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -458,7 +458,7 @@ class Aggregator:
     def remove_stale_providers(self):
         """
         Loops through all providers and check if they've gone stale. If they
-         have, write their remaining data to disk (they shouldn't have any)
+        have, write their remaining data to disk (they shouldn't have any)
         and delete them.
         """
         stale_provs = []
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -150,7 +150,7 @@ class Provider:
 
         return True
 
-    def write(self, data):
+    def save_to_block(self, data):
         """
         Saves a list of data points into blocks.
         A block will be created for any new block_name.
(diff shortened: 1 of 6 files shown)


ocs/agent/aggregator.py
```diff
@@ -149,10 +150,79 @@ def empty(self):
...
+        new_data = {}
+        for block_name, block_dict in data.items():
+            new_data[block_name] = {}
+            # rebuild block_dict
+            for k, v in block_dict.items():
+                if k == 'data':
+                    new_data[block_name]['data'] = {}
+                    for field_name, field_values in block_dict['data'].items():
+                        # replace invalid characters, and limit to 255 characters
+                        new_field_name = re.sub('[^a-zA-Z0-9_]', '', field_name)[:255]
```
jlashner, line 197: I think this replacement should also remove invalid leading characters.

BrianJKoopman: It will remove those as written:
```
>>> import re
>>> re.sub('[^a-zA-Z0-9_]', '', "!@#invalid.leading.characters")
'invalidleadingcharacters'
```

But maybe I don't follow what you're saying?

jlashner:
```
>>> import re
>>> re.sub('^a-zA-Z0-9_]', '', '1234invalid.leading')
'1234invalid.leading'
```

BrianJKoopman: Ah, good catch! Will fix.

BrianJKoopman: Alright, addressed in f566f18. Maybe some regex magic could solve it easier, but I broke down each step into somewhat easy to read parts.

BrianJKoopman:
@mhasself points out on the telecon that we should check for duplicate field names.

BrianJKoopman:
I think I'd like to make one more adjustment to this. Instead of removing spaces in field names that have them, I'd rather replace them with underscores. Other than that, this is ready for review @mhasself.

ocs/ocs_feed.py
```diff
@@ -226,30 +229,80 @@ def publish_message(self, message, timestamp=None):
...
+    def verify_data_field_string(field):
+        """There are strict rules for the characters allowed in field names.
+        This function verifies the names in a message are valid.
+
+        A valid name:
+
+        * contains only letters (a-z, A-Z; case sensitive), decimal digits (0-9), and the
+          underscore (_).
+        * begins with a letter, or with any number of underscores followed by a letter.
+        * is no more than 255 characters long.
```
mhasself, line 269: As before... this passes '' and should not.

ocs/ocs_feed.py
```diff
@@ -226,30 +229,80 @@ def publish_message(self, message, timestamp=None):
...
-                                    f"data type: {invalid_type}")
+        # single sample check
+        else:
+            if not isinstance(value, valid_types):
+                invalid_type = type(value)
+                raise TypeError("message 'data' block contains invalid " +
+                                f"data type: {invalid_type}")
+
+    @staticmethod
+    def verify_data_field_string(field):
```
mhasself, line 260: The functionality here, and in the aggregator._enforce_field_name_rules is very similar.  Perhaps a single function that serves both purposes?  That would put the rules in one place, and there would only be one function to change / test / be aware of.

The converter is an implicit checker, afterall... if standardize(x) == x then x is already compliant.  (I know there's more to it than that, for the purposes of nice error text... but still, this could all be made to live in a single place.)

ocs/agent/aggregator.py
```diff
@@ -149,10 +150,174 @@ def empty(self):
...
+                             'timestamps': [time.time()],
+                             'data': {'key1': [1],
+                                      'key2': [2]},
+                             'prefix': ''}
+                       }
+            >>> prov.save_to_block(data)
+
+            Note the block name shows up twice, once as the dict key in the
+            outer data dictionary, and again under the 'block_name' value -- in
+            this instance both the word 'test'.
```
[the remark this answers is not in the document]
jlashner: Not sure why this needs to be in the doc strings, but to answer your question the key is what's used in the function, and they should always match because the data structure is built in ocs_feed.py here: https://github.com/simonsobs/ocs/blob/74f9374bb0e5f61b51afb59a9b8a5119b25af690/ocs/ocs_feed.py#L127 
and is not set by the user.

block_name is used as both the dict key and is in Block.encoded(), so it appears twice.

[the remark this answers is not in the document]
BrianJKoopman: This wasn't clear to me when I was first trying to write a test for `save_to_block`. The messages that get passed to a feed are of the form given by `b.encoded()`, I think my assumption at the time was that `data` here followed the same form, but then I found it needed this additional block name, which is why I added it to this docstring.

ocs/agent/aggregator.py
```diff
@@ -149,10 +150,174 @@ def empty(self):
...
+
+        Returns:
+            str: A new field name that is not already in name_list
+
+        """
+        # Assume duplicate name, until proven otherwise
+        duplicate_name = True
+        name_index = 1
+
+        while duplicate_name:
```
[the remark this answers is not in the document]
BrianJKoopman: Right, that does make it more readable, will change.

Good point about the index becoming large enough to push the name over length limits. Updated to build the suffix and check its length, removing that many characters from the end of the string instead of a fixed three.

ocs/agent/aggregator.py
```diff
@@ -149,10 +150,174 @@ def empty(self):
...
+        for any new block_name.
+
+        Examples:
+            The format of data is shown in the following example:
+
+            >>> data = {'test': {'block_name': 'test',
+                             'timestamps': [time.time()],
+                             'data': {'key1': [1],
+                                      'key2': [2]},
+                             'prefix': ''}
```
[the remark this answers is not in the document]
BrianJKoopman: It's currently still a required key in blocks....

https://github.com/simonsobs/ocs/blob/74f9374bb0e5f61b51afb59a9b8a5119b25af690/ocs/agent/aggregator.py#L172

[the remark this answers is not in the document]
jlashner: Hmmm I should remove this line in the G3Timesample PR since G3Timesamples don't even have a prefix field.

ocs/ocs_feed.py
```diff
@@ -226,30 +229,80 @@ def publish_message(self, message, timestamp=None):
...
+
+        # Similar to check_invalid, search for non letter characters
+        # Leading ^ matches the start of string, so following numbers are valid
+        check_start = re.compile('^[^a-zA-Z]')
+
+        # check for invalid characters
+        result = check_invalid.search(field)
+        if result:
+            raise ValueError("message 'data' block contains a key with the " +
+                             f"invalid charcter '{result.group(0)}'. "
```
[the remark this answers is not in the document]
BrianJKoopman: Fixed.

ocs/agent/aggregator.py
```diff
@@ -149,10 +150,174 @@ def empty(self):
...
+        name_index = 1
+
+        while duplicate_name:
+            if field_name not in name_list:
+                duplicate_name = False
+            else:
+                if len(field_name) < 252:
+                    field_name += f'_{name_index:02}'
+                else:
+                    field_name[:-3] += f'_{name_index:02}'
```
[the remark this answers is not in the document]
BrianJKoopman: Right, fixed and test case added.

ocs/ocs_feed.py
BrianJKoopman: I suppose. I think this came about from thinking of the Feed system first, with the thought that if everything is updated simultaneously (which is now the plan) it'll be impossible for invalid field names to get to the aggregator. This `Feed.verify_data_field_string()` method is used in the aggregator `_verify_provider_data()` method, which checks field names in the aggregator and marks them as needing to be modified before writing to disk.

A merged function would have to track the various reasons a field name was invalid to report errors in the instance of the Feed where we want the Agent to exit with a raised exception, vs in the aggregator where we don't want that behavior. 

I do think that even if this was changed we'd want to be testing it via test calls to the feed `Feed.publish_message()` and aggregator `Provider.save_to_block()`, so I'm not sure it actually reduces the number of tests run.

Admittedly, it was this separation that lead to missing several test cases that were caught by this PR, but now that we have it mostly working I'm somewhat resistant to rewriting it. I can do it if you feel strongly about it though. Do we anticipate changing these rules much in the future? (Such that it would be inconvenient to have the rule check and enforcement separate.)

ocs/agent/aggregator.py
```diff
@@ -149,10 +150,174 @@ def empty(self):
...
+    @staticmethod
+    def _enforce_field_name_rules(field_name):
+        """Enforce naming rules for field names.
+
+        A valid name:
+
+        * contains only letters (a-z, A-Z; case sensitive), decimal digits (0-9), and the
+          underscore (_).
+        * begins with a letter, or with any number of underscores followed by a letter.
+        * is no more than 255 characters long.
```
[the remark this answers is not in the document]
BrianJKoopman: Right, how do we want to handle a blank field though? I'm thinking I'd just insert "unknown_field" as the field name.

ocs/agent/aggregator.py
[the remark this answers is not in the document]
mhasself: I believe the latest version is still a little off -- instead of hard-coded 252 you want 255 - suf_len.

ocs/ocs_feed.py
mhasself: I don't think this was addressed in yesterday's commits.

BrianJKoopman: Yeah, haven't finished addressing everything yet, just wanted to push my latest work, sorry if that gave the impression I was finished up. Can re-request review when that's true.

ocs/agent/aggregator.py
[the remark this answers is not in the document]
BrianJKoopman: Thanks, there seems to always be an edge case I haven't considered. Added a catch if rule enforcement returns an empty string in c46f95c.

ocs/ocs_feed.py
BrianJKoopman: This should be resolved (and tested) by c46f95c.

ocs/agent/aggregator.py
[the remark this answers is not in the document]
BrianJKoopman: Right, fixed.

ocs/agent/aggregator.py
```diff
@@ -257,12 +257,12 @@ def _check_for_duplicate_names(field_name, name_list):
...
-                field_name += f'_{name_index:02}'
+            suffix = f'_{name_index:02}'
+            suf_len = len(suffix)
+            if len(field_name) < (255 - suf_len):
+                field_name += suffix
             else:
-                suffix = f'_{name_index:02}'
-                suf_len = len(suffix)
-                field_name = field_name[:-1*suf_len] + f'_{name_index:02}'
+                field_name = field_name[:-1*suf_len] + suffix
```
mhasself, line 265: This entire if block could be replaced by `field_name = field_name[:255-suf_len] + suffix`... but I'm being pedantic about an extreme edge case so it's probably safe to ignore me...

PR #139, opened 2020-06-04
Bump so3g version

BrianJKoopman:
This introduces support for HK format v1 to the ocs image.

(This is done in two commits just because of an old testing commit I pushed to this branch. Will squash and merge when tests on the latest pass.)


mhasself:
Good time to tag so3g v0.0.8?

PR #147, opened 2020-07-10
checkdata invalid fields update

BrianJKoopman:
This adds a feature to the `checkdata` script that will flag invalid field names by displaying them in yellow in the verbose output.

For example:
image

It also contains some changes for dynamically handling the field name column width.


BrianJKoopman:
I don't think there's anything controversial here, so I'll plan to merge soon unless comments come in. Functionality here is helpful in testing https://github.com/simonsobs/socs/pull/87.

PR #150, opened 2020-07-16
setup.py: add ocs.agent submodule

mhasself:


mhasself:
Without this, the agent submodule is not installed and the aggregator Agent is not able to "import ocs.agent" unless running from the source tree.

BrianJKoopman:
Can we import `find_packages` from setuptools and do a `packages = find_packages()`? That'll avoid forgetting to add a package in the future. socs does this.

BrianJKoopman:
Cool, I dunno if I jumped the start on requesting review, but it seemed like a quick change. Approved!

mhasself:
Closing to PR against develop branch.

PR #154, opened 2020-07-20
Influx publisher updates

### LINKED ISSUE #113: Handle missing influxdb on init in InfluxDB Publisher
If the InfluxDB isn't running when you start the publisher it won't be able to connect and will crash. While we handle the DB disappearing after startup, we don't handle the case of a missing DB pre-startup. We suggest Influx is run separate from other ocs containers (in a different docker-compose file) and kept online at all times, if a user configures both services in a single docker-compose likely the publisher comes up before influx and thus will crash immediately.

BrianJKoopman:
## Description
Address two items related to the InfluxDB Publisher:
1. Fix #113 
2. Allow user configurable database name.

## Motivation and Context

Previously we defaulted to the hard coded 'ocs_feeds' database name. There has only ever been one database, and setting the name wasn't initially important. Enabling user configurable database names allows users to re-upload data from .g3 files to a separate database instance, for instance if some reprocessing of data on disk needs to take place (see https://github.com/simonsobs/socs/pull/87).

## How Has This Been Tested?

An isolated set of Docker containers containing core OCS components was run, first with the InfluxDB offline, checking for the reconnection attempts. Bringing the InfluxDB online then allows for successful reconnection and proceeding data publication (tested with an LS240 simulator.)

The same set of containers was then configured to publish data to a separate database, not the default 'ocs_feeds'. The new database was created and data successfully published to it.

The stack was brought up again and verified data published after fixing the flake8 errors as well.

## Types of changes

## Checklist:

### DIFF
--- a/agents/influxdb_publisher/influxdb_publisher.py
+++ b/agents/influxdb_publisher/influxdb_publisher.py
@@ -39,6 +39,8 @@ class Publisher:
             A thread-safe queue of (data, feed) pairs.
         host (str):
             host for InfluxDB instance.
+        database (str):
+            database name within InfluxDB to publish to
         port (int, optional):
             port for InfluxDB instance, defaults to 8086.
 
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
Small bug fix and a helpful feature addition for handling the field renaming/update to HK v1 going on. Going to merge so I can use the `-dev` image. Please open an issue with any concerns if you have them soon (before we prepare for the v0.7.0 release.)

PR #158, opened 2020-07-30
Finish moving to GitHub Actions

### LINKED ISSUE #143: Switch CI Pipeline to GitHub Actions
GitHub Actions looks like it provides better control over when a job is run. We've wanted to build and push docker images on releases/tags only for a while now, but the solution for that hasn't seemed simple to me. It looks fairly straight forward to do in GitHub Actions, so I don't see a reason not to switch.

BrianJKoopman:
## Description

Finalize the move to GitHub Actions. We also introduce building the docs to the CI pipeline so that we know if they break during PRs, etc. The "build" badge now points to the `develop` branch workflow.

resolves #143 

## Motivation and Context

Using the GitHub Action for coveralls hadn't worked when I tried it initially, related to the format it's expecting and what format we were outputting to. We go back to calling `coveralls` directly, with a github token required for publishing to coveralls from GitHub actions. This solves the coveralls problem, and allows us to move totally off of Travis.

Documentation builds are also a common breakage, so we start testing those in the pipeline.

I went back and forth on how best to setup the pipeline to ensure testing happens, coverage is published, and docker images built and pushed to Docker Hub. There are three workflows, and probably we're running the tests (pytest and building the docs) more often than we have to, but they don't take long to run and I'd rather err on running them too often than not enough. The summary of the pipeline is:
* Run pytest and the docs build every time we do something
* Report code coverage on all pushes (this is separate per branch)
* Build `-dev` docker images only on pushes to develop (i.e. after a PR is merged to develop)
* Build "official" (i.e. v0.1.0) docker images only on releases.

## How Has This Been Tested?

The pytest workflow runs on all pushes, so that's been tested several times in this PR alone as I tweaked things. 

## Types of changes

## Checklist:


BrianJKoopman:
Squashing and merging this in before the 0.7.0 release, as that would be a good verification of the workflows. there are no actual code changes, just CI changes (with maybe the exception of adding the `sphinx_rtd_theme` module to the requirements file, but I don't expect that to affect anything.)

PR #163, opened 2020-08-26
OCS Web control panel system for Agents

mhasself:
Significant advancements to the OCS Web system.  monitor.html now has a tabbed interface where panels can be opened for controlling and monitoring each Agent.  A "generic" control panel can be opened, or a panel matched to the specific Agent Class can be used in order to provide operation-specific parameters.

This has significant documentation describing how to develop a new custom panel.

## Description

This includes simple "specialized" panels for the Aggregator and FakeData agents, as well as the Generic panel.  Simple panels for other agents can be made by copying the (fairly compact) examples these provide.  

There are some changes to ocs/ocs_agent.py and site_config.py, in order to store and expose the Agent's class name (e.g. 'FakeDataAgent') to the world.

## Motivation and Context

I don't believe there's an issue for this, but it's been the plan for a long time and has been re-re-re-requested recently.

## How Has This Been Tested?

This has been tested with a combined docker+host OCS setup, using recent versions of Firefox and Chrome.

## Types of changes

## Checklist:


BrianJKoopman (approved):
This is great! Couple of small comments inline below. Other misc. thing I came across/questions I had:
* In testing I found I can start a delay task, but can't abort mid-delay.
* Is the name `delay_box` special in any way, or are we just trying to avoid a collision with the name `delay`. If I can make a UI suggestion, can we have the "Suggestions" drop down populate the "Delay (s)" field with the selected value? That way the "Delay" field is the definitive place for the delay parameter, and there is no potential confusion between selecting a suggestion and entering a value.
* `set_boxes()` seems to be missing from the OcsUiHelper Method Reference

I should note that I've looked at the Python changes here, and the FakeDataAgent.js, but not most of the other .js or .html/.css.

I don't want to hold up moving forward on this, so maybe if the UI suggestion is good we should open an issue for it. Similarly for the abort functionality, I assume that should be working?

mhasself:
Thanks -- pushed fixes and will merge soon if there are no further objections.

- "Abort" is part of OCS API but there is no way for Agents to handle it currently... so I've commented it out of the UI for now.
- Addressed your documentation points.
- You're right that the dropdown is silly, but perhaps we can just plan to delete it when there's a better Operation in FakeData Agent that can use it as a UI element.  Having the element change another element's value is a more advanced example.

Issue #92, opened 2019-11-08
Lots of required packages


jlashner:
`pip install -r requirements.txt` Currently installs a lot of packages
```
Installing collected packages: lmdb, setuptools, zope.interface, hyperframe, hpack, h2, six, pycparser, cffi, cryptography, psutil, constantly, txaio, autobahn, priority, bitstring, pygments, pyyaml, bcrypt, netaddr, setproctitle, pyqrcode, sortedcontainers, pytrie, pyopenssl, PyHamcrest, incremental, attrs, Automat, idna, hyperlink, twisted, urllib3, certifi, chardet, requests, treq, mistune, werkzeug, MarkupSafe, jinja2, click, pyasn1, passlib, txtorcon, cbor2, pynacl, numpy, zlmdb, u-msgpack-python, sdnotify, py-ubjson, pyasn1-modules, service-identity, cbor, argh, pathtools, watchdog, crossbar, simplejson, greenlet, dnspython, monotonic, eventlet, gevent, wampy, pyparsing, packaging, deprecation, msgpack, pyserial, pytz, babel, imagesize, sphinxcontrib-devhelp, sphinxcontrib-applehelp, sphinxcontrib-qthelp, snowballstemmer, alabaster, sphinxcontrib-serializinghtml, docutils, sphinxcontrib-jsmath, sphinxcontrib-htmlhelp, sphinx, sphinx-argparse, atomicwrites, more-itertools, zipp, importlib-metadata, pluggy, py, wcwidth, pytest, coverage, pytest-cov, docopt, coveralls
[...]

PR #166, opened 2020-09-11
Create ocs-crossbar Dockerfile and remove sisock refs

### LINKED ISSUE #135: Redraw boundary between ocs and sisock
sisock is currently woven into ocs installations:
- the base crossbar image includes the sisock hub -- this is an unnecessary modularity fail; the hub can be added on as an Agent (or similar) if needed.
- lab crossbar configs include SSL configuration w/secrets, complicating git use; let's just use simpler crossbar configs.

It seems unlikely that we will run OCS to control an observatory and also allow external (on internet, requiring SSL) WAMP connections to pull data.  Even if we did, it's probably better to set up SSL using the webserver proxy features.

The current best solution for live viewing of data makes use of grafana and influxdb.  The most profitable role for sisock to focus on will be the efficient delivery of multi-resolution archival data.

BrianJKoopman:
## Description

"sisock-crossbar" is the old crossbar server provided by the sisock repo.
Per #135, we want to redraw the separation between sisock and ocs. This creates
a Dockerfile from which we build a crossbar server Docker image with a sensible
default ocs-centric crossbar configuration already in the image. We also rename
all references to "sisock-crossbar" to simply just "crossbar" in the docs.

## Motivation and Context

This has been an open issue for a while now, but recently I was going through the dependencies required by OCS in the context of https://github.com/simonsobs/spt3g-docker/issues/1 and I was finding the crossbar package installation requires a specific (and now kind of dated) numpy version. Splitting this requirement out to just the crossbar container will prevent any conflicts in Agents where newer versions of numpy might be preferred, like the aggregator.

I'm planning on handling splitting the requirements.txt file, including breaking out the crossbar requirement which pins the numpy version, in a separate PR while addressing #92.

closes #135

## How Has This Been Tested?

I've run a small dev setup on my laptop in Docker containers consisting of the crossbar server, an aggregator, ls240 Agent, and ls240 simulator. Things connect as expected. 

## Types of changes

## Checklist:


mhasself (approved):
Great...

I took a stab at testing this; one glitch I encountered was needing to override the --site* args for the aggregator.  Should the default args in agents/*/Docker be changed now too?

BrianJKoopman:
> Great...
> 
> I took a stab at testing this; one glitch I encountered was needing to override the --site* args for the aggregator. Should the default args in agents/*/Docker be changed now too?

Mhm, good point. Updated in a7c5417 with s/sisock-crossbar/crossbar. This'll have to be announced with the next version of OCS, as this is going to need changing in every deployed setup, since everyone uses "sisock-crossbar" as their crossbar server.

PR #169, opened 2020-09-15
site_config: use a string comparison instead of identification

mhasself:
setup.py is catching this bad usage.

It's a trivial fix, speaks for itself...


BrianJKoopman (approved):
Great, thanks!

Issue #152, opened 2020-07-17
Provide helper library for regular sampling rates


BrianJKoopman:
We discussed on the 2020-07-16 DAQ call adding some code to assist developers achieve regular sampling rates in their Agents. Currently many Agents simply sleep for some set time, almost ensuring a slower sampling rate then desired. Groups have also seen slight variations in the time between points.

The code we add should assist in having the Agent wait the appropriate amount of time. It might be nice if this could be aligned to the nearest second where reasonable, i.e. if the desired sampling rate is 4 Hz, taking data at 9:00:15.0, 9:00:15.25, 9:00:15.5, and 9:00:15.75 (to use the time on my current clock as an example...)

BrianJKoopman:
closed by #178

PR #178, opened 2020-10-06
Pacemaker class

jlashner:
Adds a "Pacemaker" class to OCS to help Agents maintain regular sampling frequencies.

## Description

This PR adds a new helper class to OCS to help agents such as the Lakeshore372 maintain a regular sampling rate despite having data acquisition functions that take varying amounts of time. The Pacemaker also lets agents "quantize" their samples for sampling rates that are whole numbers, forcing samples to land on a temporal grid starting on the second, which will help synchronize data acquisition across agents. 

## Motivation and Context

Closes issue #152.

## How Has This Been Tested?

Added a test function that tests sampling regularity in a loop for  few seconds. Have not implemented into agents yet.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/ocs_twisted.py
+++ b/ocs/ocs_twisted.py
@@ -1,6 +1,8 @@
 import threading
 from contextlib import contextmanager
 import time
+from autobahn.twisted.util import sleep as dsleep
+from twisted.internet.defer import inlineCallbacks
 
 
 class TimeoutLock:
@@ -156,7 +158,7 @@ class Pacemaker:
     Here is an example of how the Pacemaker can be used keep a 3 Hz quantized
     sample rate::
 
-        pm = Pacemaker(10, quantize=True)
+        pm = Pacemaker(3, quantize=True)
         take_data = True:
         while take_data:
             pm.sleep()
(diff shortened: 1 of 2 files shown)


mhasself (changes requested):
This implementation uses time.sleep, and thus can't run in the twisted reactor.  Perhaps add a dsleep method that does the same thing but yields from dsleep?

BrianJKoopman (changes requested):
## Tests
For completeness I think a couple more tests could be added. One that tests for invalid inputs when `quantize=True`, i.e.:
```
def test_non_integer_quantization():
    """Trying to quantize with a non-integer sampling frequency should raise an
    error.

    """
    with pytest.raises(ValueError):
        Pacemaker(5.5, quantize=True)
```
The other, which verifies the pacemaker works when `quantize=False`, the distinction maybe being that you include that first diff you omit in the current test due to the quantization.

jlashner:
Implemented suggestions. I also tested the dsleep method with the following test:
```
from ocs.ocs_twisted import Pacemaker
from twisted.internet.defer import inlineCallbacks
from twisted.internet import reactor
import time

@inlineCallbacks
def test_dsleep():
    pm  = Pacemaker(3, quantize=True)
    for _ in range(10):
        yield pm.dsleep()
        print(time.time())

reactor.callWhenRunning(test_dsleep)
reactor.run()
```
However I'm not sure if we have the framework to test twisted functions with pytest yet.

BrianJKoopman (approved):
Looks good, thanks for the added tests.

On the topic of testing twisted functions. Twisted has their own test runner capable of this, called Trial, which is meant to stand in for the python `unittest` framework. However, there's also this `pytest-twisted` plugin, which might be worth checking out: https://pypi.org/project/pytest-twisted/

PR #179, opened 2020-10-08
Remove empty blocks' block_name from frame

BrianJKoopman:
## Description

If a block has no timestamps, don't include the "block_name" in the frame. Add a test to verify this behavior.

## Motivation and Context

If a previously published block wasn't published during a frame, the block's
name would end up in the frame's "block_names" list anyway. We now don't include it if
the block is empty, i.e. if there are no timestamps.

This was found when UCSD started using the M1000 Agent. It contains a very sparsely sampled field and associated block. They were unable to open saved files, which led to finding invalid frames being saved with `len(block_names) != len(blocks)`. 

## How Has This Been Tested?

New test, `test_sparsely_sampled_block`, added to `test_aggregator.py`.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -445,7 +445,10 @@ class Provider:
             frame['blocks'].append(m)
             block_names.append(block_name)
 
-        frame['block_names'] = core.G3VectorString(block_names)
+        if 'block_names' in frame:
+            frame['block_names'].extend(block_names)
+        else:
+            frame['block_names'] = core.G3VectorString(block_names)
 
         if clear:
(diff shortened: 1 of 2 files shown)


jlashner (approved):
Looks good!

mhasself (approved):
Thanks for the quick fix.

Note that so3g.hk.HKScanner does have a check for this schema violation.  It might be worth integrating that into your tests or verification process.

https://github.com/simonsobs/so3g/blob/456feb86c3ed4aa26c28ae3fe47f7dcaefc51ed3/python/hk/scanner.py#L140

PR #184, opened 2020-10-23
Build ocs-crossbar image from official crossbario/crossbar image

BrianJKoopman:
## Description

Instead of building our own crossbar image by installing crossbar on an ocs base image, use the [official crossbario/crossbar](https://hub.docker.com/r/crossbario/crossbar) image from Dockerhub. This bundles a default config generated by ocsbow into the image, which should mostly eliminate the need for test institutions to generate and mount a crossbar config.

Also update the documentation quite a bit to describe this change and how to mount a new config into the container/rebuild if needed. I got a little carried away to documentation work, as this also includes updates to instructions that describe how to split out various components to run separate from the ocs docker-compose.yml file so we can avoid always bringing down services like influxdb/crossbar/grafana.

## Motivation and Context

The build of crossbar we were making was broken. Rather than fix it I thought it best to switch to the official image, or at least one based on the official image. This uses a `crossbar` user internally, so has some small hurdles if mounting a config file, but the way to jump over those is described in the new docs.

## How Has This Been Tested?

I have built and run the new crossbar image locally in a small test environment.

## Types of changes

## Checklist:

### DIFF
--- a/docs/user/crossbar_config.rst
+++ b/docs/user/crossbar_config.rst
@@ -18,102 +18,21 @@ the interface to the crossbar server.
 
 Example Config
 --------------
-Here is the default OCS crossbar config that is bundled into
-``simonsobs/ocs-crossbar``: 
+An example of the default OCS crossbar config that is bundled into
+``simonsobs/ocs-crossbar`` can be found in the repository at
+`ocs/docker/crossbar/config.json`_. This is based on the template in
+`ocs/ocs/support/crossbar_config.json`_.
 
-.. code-block:: json
+The unique parts of this to OCS are the realm name, "test_realm", defined
+roles of "iocs_agent" and "iocs_controller, and "address_root" of
+"observatory.". Additionally, we run on port 8001.
 
-    {
-        "version": 2,
-        "workers": [
-            {   
-                "type": "router",
-                "options": {
-                    "pythonpath": [
-                        ".."
-                    ]   
-                },  
-                "realms": [
-                    {   
-                        "name": "test_realm",
-                        "roles": [
-                            {   
-                                "name": "iocs_agent",
-                                "permissions": [
-                                    {   
-                                        "uri": "observatory.",
-                                        "match": "prefix",
-                                        "allow": {
-                                            "call": true,
-                                            "register": true,
-                                            "publish": true,
-                                            "subscribe": true
-                                        },  
-                                        "disclose": {
-                                            "caller": false,
-                                            "publisher": false
-                                        },  
-                                        "cache": true
-                                    }   
-                                ]   
-                            },  
-                            {   
-                                "name": "iocs_controller",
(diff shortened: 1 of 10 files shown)


docs/user/crossbar_config.rst
```diff
@@ -0,0 +1,120 @@
...
+For further details on crossbar server configuration, see the crossbar `Router
+Configuration`_ page.
+
+.. _`ocs/docker/crossbar/config.json`: https://github.com/simonsobs/ocs/blob/develop/docker/crossbar/config.json
+.. _`ocs/ocs/support/crossbar_config.json`: https://github.com/simonsobs/ocs/blob/develop/ocs/support/crossbar_config.json
+.. _`Router Configuration`: https://crossbar.io/docs/Router-Configuration/
+
+Generating a New Config File
+----------------------------
+``ocsbow`` can be used to generate a default configuation file, based on
```
mhasself, line 39: typo

docs/user/crossbar_config.rst
```diff
@@ -0,0 +1,120 @@
...
+
+You should then update your configuration to use the new, local,
+``ocs-crossbar`` image.
+
+Bind Mounting the Configuration
+-------------------------------
+To instead mount the new configuration into the pre-built image, first chown
+your file to be owned by user and group 242 (the default crossbar UID/GID),
+then mount it appropriately in your docker-compsose file. Here we assume you
+put the configuration in the direcitory ``./dot_crossbar/``::
```
mhasself, line 107: typo

docs/user/crossbar_config.rst
```diff
@@ -0,0 +1,120 @@
...
+
+    $ chown -R 242:242 dot_crossbar/
```
mhasself, line 109: Can we explore other ways of doing this?  It looks like there are ways to map a host system uid/gid to a container uid/gid.  User/group 242 on the host system could be anything, and it's odd to have files lying around with bizarre ids like this.

https://linuxnatives.net/2019/align-user-ids-inside-and-outside-docker-with-subuser-mapping

mhasself (approved):
Thanks for this.

Fixing the chown thing shouldn't hold this up but I think we should figure out if there's a better way to handle permissions on mounted volumes going forward.

docs/user/crossbar_config.rst
BrianJKoopman: I'll definitely look into this. I wasn't crazy about the chown solution. That was part of the reason I avoided using this base image in the first place. It's worth figuring out though, so we can run as a non-root user within the containers.

BrianJKoopman:
> Thanks for this.
> 
> Fixing the chown thing shouldn't hold this up but I think we should figure out if there's a better way to handle permissions on mounted volumes going forward.

Great, I'll merge now then and open an issue about the permissions problem.

PR #183, opened 2020-10-22
Ignore and log invalid kwargs passed to Providers

BrianJKoopman:
## Description

Adds a ``**kwargs`` parameter to the `Providers` class, allowing us to ignore invalid kwargs passed via the `agg_params` dict defined in Agents and passed to `OCSAgent.register_feeds()`.

Note that this doesn't provide the Agent writer any real feedback on the fact they've included invalid kwargs, but does log it to the Aggregator Agent logs as a warning.

## Motivation and Context

fixes #148

## How Has This Been Tested?

A unit test has been added that creates a provider with invalid kwargs. This would fail without the new ``**kwargs`` parameter.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] Unless I am preparing a release, I have opened this PR onto the `develop` branch.

PR #187, opened 2020-11-07
Added feed exclusion options

jlashner:
Adds the `exclude_aggregator` and `exclude_influx` aggregation parameters, which will keep data from being published to g3 and influx respectively.

## Description

Added a check in the aggregator / influx feed callbacks that check for the `exclude_aggregator` or `exclude_influx` option, and will toss the data if set to `True`.

## Motivation and Context

This allows for "influx only" or "aggregator only" feeds. Right now it's needed to allows feeds with higher data rates to bypass the influx publisher while still being recorded in 3g.

## How Has This Been Tested?

Tested locally with the `fake_data` agent. Checked to see that no data is published to influx when the parameter is set.

## Types of changes

## Checklist:

### DIFF
--- a/agents/influxdb_publisher/influxdb_publisher.py
+++ b/agents/influxdb_publisher/influxdb_publisher.py
@@ -90,6 +90,8 @@ class Publisher:
         """
         while not self.incoming_data.empty():
             data, feed = self.incoming_data.get()
+            if feed['agg_params'].get('exclude_influx', False):
+                return
 
             LOG.debug("Pulling data from queue.")
 
--- a/agents/influxdb_publisher/influxdb_publisher.py
+++ b/agents/influxdb_publisher/influxdb_publisher.py
@@ -91,7 +91,7 @@ class Publisher:
         while not self.incoming_data.empty():
             data, feed = self.incoming_data.get()
             if feed['agg_params'].get('exclude_influx', False):
-                return
+                continue
 
             LOG.debug("Pulling data from queue.")
 
--- a/ocs/agent/aggregator.py
+++ b/ocs/agent/aggregator.py
@@ -152,7 +152,7 @@ class Provider:
             txaio logger
 
     """
-    def __init__(self, address, sessid, prov_id, frame_length=5*60, fresh_time=3*60, **kwargs):
+    def __init__(self, address, sessid, prov_id, frame_length=5*60, fresh_time=3*60):
         self.address = address
         self.sessid = sessid
         self.frame_length = frame_length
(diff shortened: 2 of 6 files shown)


agents/influxdb_publisher/influxdb_publisher.py
```diff
@@ -90,6 +90,8 @@ def process_incoming_data(self):
         """
         while not self.incoming_data.empty():
             data, feed = self.incoming_data.get()
+            if feed['agg_params'].get('exclude_influx', False):
+                return
```
mhasself, line 94: Should this should be "continue", not "return"?  You want to empty the queue on this call and not rely on call rate > block arrival rate.  (I might have read the code too quickly, but please check.)

ocs/agent/aggregator.py
```diff
@@ -617,6 +617,9 @@ def process_incoming_data(self):
 
             data, feed = self.incoming_data.get()
 
+            if feed['agg_params'].get('exclude_aggregator', False):
+                return
```
mhasself, line 621: "continue" again, I suspect.

mhasself (changes requested):
See inline question.

While Brian tests functionality... some documentation suggestions:

- The agg_params are now described in two separate docstrings... perhaps the register_feed docstring should refer to the version in Feed, at least for the agg_params arguments?
- These new controls aren't mentioned in (as far as I can tell) the salient section of the docs, [here](https://ocs.readthedocs.io/en/develop/developer/feeds.html#recorded-feed-registration).  Provide links from there to the auto-documentation of Feed?

agents/influxdb_publisher/influxdb_publisher.py
jlashner: Good catch, thanks! Should definitely be continue in both cases.

BrianJKoopman (changes requested):
This works for me. I tested both excluding from the aggregator and from the InfluxDB.

One small thing -- `exclude_influx` gets flagged as an unexpected kwarg by the Provider class due to the changes in #183. This'll cause it to show up in the logs (actually, there's a bug in that log.warn statement, it always shows up with an empty list...). I think we should do something to avoid this, since it is now expected.

jlashner:
Good point, I think it's probably a good idea to separate the "agg_params" dictionary from the Provider class kwargs and remove that catch all, since it doesn't really make sense to input it directly anyway the way we were doing before. I've implemented it, but feel free to veto the change if you disagree.

BrianJKoopman:
> [I]t doesn't really make sense to input it directly anyway the way we were doing before.

Agreed, the update looks good. Can you remove the failing `test_arbitrary_args_to_provider` test too? That's no longer valid with the Provider change.

BrianJKoopman (approved):

PR #188, opened 2020-11-10
Operation Monitoring

jlashner:
This PR adds the `agent_operations` feed for the registry agent which allows for the monitoring of individual agent operations, and makes it possible to set alerts for failed processes in grafana.

## Description

The agent heartbeat now publishes a dict of operation codes for the agent which can be used to determine the state of each operation that the agent can run. The operation codes are the following:
- `NONE`: If an operation has never been run
- `STARTING`: If an operation currently has the 'starting' status
- `RUNNING`: If the operation has the 'running' status
- `STOPPING`: If the operation has the 'stopping' status
- `SUCCEEDED`: If the operation has finished with `success=True`
- `FAILED`: If the operation has finished with `success=False`
- `EXPIRED`: If the agent who owns the operation is no longer running (set by the registry).

## Motivation and Context

This lets you make cool grafana panels that track the status of all operations:
Screen Shot 2020-11-09 at 8 39 32 PM

It also makes it possible to setup grafana alerts for individual operations instead of having to monitor docker updates with telegraf.

## How Has This Been Tested?

This has been tested by running agents locally.

## Types of changes

## Checklist:


agents/registry/registry.py
```diff
@@ -76,13 +84,22 @@ def __init__(self, agent):
             options={'match': 'wildcard'}
         )
 
+        agg_params = {
+            'frame_length': 30*60,
```
BrianJKoopman, line 88: This is the longest frame length in any Agent. Did we make a formal recommendation of 10 minute frames? All other Agents either have 10 or 1 minute frames.

BrianJKoopman (commented):
This is great, I'm sure this functionality will see a lot of use. A couple of small suggestions inline, mostly to suggest documentation. The screenshot of the panel might even be good to include.

What'd you mean by "... instead of having to monitor docker updates with telegraf." Are you referring to the logs?

ocs/ocs_agent.py
```diff
@@ -739,6 +747,17 @@ def encoded(self):
 
 SESSION_STATUS_CODES = [None, 'starting', 'running', 'stopping', 'done']
 
+
+class OpCode(Enum):
+    NONE = 1
```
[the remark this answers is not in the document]
jlashner: I started at one because the [Enum documentation](https://docs.python.org/3/library/enum.html) always starts at one, so I thought there might be something weird with enum values of zero. But if you'd prefer I can start at 0.

agents/registry/registry.py
jlashner: I'll change this to 1 min.

ocs/ocs_agent.py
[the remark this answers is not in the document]
BrianJKoopman: Ah, I see why they do this now:

> The reason for defaulting to 1 as the starting number and not 0 is that 0 is False in a boolean sense, but enum members all evaluate to True.

I don't think it would cause us any problems, but probably best to leave it as is.

ocs/ocs_agent.py
```diff
@@ -739,6 +747,17 @@ def encoded(self):
...
+    STARTING = 2
+    RUNNING = 3
+    STOPPING = 4
+    SUCCEEDED = 5
+    FAILED = 6
+    EXPIRED = 7
```
mhasself, line 758: The Enum is potentially of interest to consumers that don't have twisted available... and thus might not want to import ocs_agent.  Can it be moved to "base.py", which is sort of like the header file for ocs?

mhasself: I acknowledge that I did not do this with 'starting', 'running', ...  I should have!

mhasself (approved):
Cool!

This approval is contingent on Brian being happy with the documentation.

jlashner:
Alright, I added a docs section and implemented all suggestions. I also added field name validation to the registry fields before adding them to the data block because I realized long or improper agent instance-id's would result in improper field names in the `agent_operations` feed.

Let me know if there's anything else I should add before merging!

mhasself (approved):

agents/registry/registry.py
```diff
@@ -143,17 +146,21 @@ def main(self, session: ocs_agent.OpSession, params=None):
...
+                        'data': {}}
+                for op_name, op_code in agent.op_codes.items():
+                    field = f'{addr}_{op_name}'
+                    field = field.replace('.', '_')
+                    field = field.replace('-', '_')
+                    try:
+                        Feed.verify_data_field_string(field)
+                    except ValueError as e:
+                        self.log.warn(f"Improper field name: {field}\n{e}")
+                        continue
```
BrianJKoopman, line 160: So agent address + op_name fields that are invalid are simply dropped? This is tricky to handle I suppose because an Agent could come online at any time with an address that could make this invalid. What's likely to happen is the Agent is brought up, and no operation status shows up for the user when expected, with no feedback on their end (since it's in the Registry.) This is the same problem we had with the Aggregator generating errors that would go unnoticed when new Agents would come online. Checking this combination of addr + op_name on the Agent heartbeat side would provide the user starting up the Agent feedback and prevent an invalid combo from getting this far.

The field name rules could also be enforced for invalid names. The Provider class does this for instances where older Agents are running on the newer Aggregator: https://github.com/simonsobs/ocs/blob/bba8a3f3378bee0eb9a6558f848579366b0d4570/ocs/agent/aggregator.py#L227

BrianJKoopman (commented):
That's a good point about the field naming potentially becoming invalid. I commented on this inline below. I think moving this check closer to the heartbeat could provide the Agent user more feedback.

agents/registry/registry.py
jlashner: I think using `_enforce_field_name_rules` is a good solution, I'll do that instead of adding registry-specific warnings to the `register_task` and `register_process` OCSAgent functions.

jlashner:
Fixed Brian's last comments. Added `Provider._enforce_feed_name_rules` as a final step to make sure fields are monitored instead of dropped.

BrianJKoopman (approved):
Changes look good, thanks for those! Maybe longer term that method marked as hidden on the Provider should be moved somewhere more general or made not hidden, but this works for now.

PR #194, opened 2021-02-11
Drop wampy module as dependency

BrianJKoopman:
## Description

With the new pip resolver in versions >=20.3 wampy is causing some issues with dependency resolution. It has hard requirements for [gevent](https://pypi.org/project/gevent/) and [greenlet](https://pypi.org/project/greenlet/), which conflict with other module dependencies. We don't really use wampy clients anywhere, so it was decided we could drop the dependency. This'll also clear up some need to avoid the wampy plugin when running tests.

Note: This is based on the [ubuntu-20.04](https://github.com/simonsobs/ocs/tree/ubuntu-20.04) branch, else things wouldn't have built properly. That should be merged first.

## Motivation and Context

See description.

## How Has This Been Tested?

Unit tests pass. Integration tests pass (where expected anyway.)

## Types of changes

## Checklist:


docs/developer/architecture.rst
```diff
@@ -85,12 +85,5 @@ environments.
...
-asynchronous Control Clients, such as example_ctrl.py.  However, one
-can write Control Clients instead using the simpler, non-asynchronous
-WAMP library called ``wampy``; see below.
-
-wampy
-`````
-A simple, blocking Control Client can be written in python using only
-the ``wampy`` library. Such a client uses the module ocs.client_wampy.
+These are required to run an OCS Agent. They are also required to run
+asynchronous Control Clients, such as example_ctrl.py.
```
mhasself, line 89: This is fine but it might bear mentioning that "However, simple clients can be written in the usual Python synchronous framework, without any need for twisted or autobahn."

mhasself (approved):
Minor comments -- thanks!

tests/test_site_config.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class TestGetControlClient:
+    """Test site_config.get_control_client().
+
+    """
+    def test_none_client_type(self):
+        """None should assume 'http' appropriately."""
+        mock_site = MagicMock()
+        get_control_client('test', site=mock_site)
```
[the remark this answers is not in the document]
BrianJKoopman: Right, I think I wrote this before setting `client_type='http'` in the arguments. But yeah, was just making sure a test hit the line that checked when `client_type` was `None`. Removed since it's redundant now.

As for `site.hub.data` below, yeah, I was just trying to make sure if anyone has any clients that specify type cases are handled correctly. Totally crashes if site.hub.data can't be checked, so I used the Mock.

docs/developer/architecture.rst
BrianJKoopman: Sentence added, thanks!

PR #199, opened 2021-04-02
Remove deprecated reparse_args usage

BrianJKoopman:
## Description

`site_config.reparse_args()` has been deprecated for a while now, but is still very widely used in socs Agents. This removes it from the only core agent still using it, from one of the examples, and perhaps mostly importantly from the Agent developer documentation.

The one place I haven't removed it from is from ocsbow. I was finding in removing it from some socs Agents that if the Agent still called this line:
```
parser = site_config.add_arguments()
```
to initialize the parser, when `site_config.parse_args()` was later called (which internally calls `site_config.add_arguments()`) we get a conflict error from argparse. Maybe we can discuss how this should be handled here? This PR at least updates examples that might get copied as new Agents are developed. Then an issue can be created if we don't end up handling how `add_arguments()` interacts with things.

## Motivation and Context

It's common for new Agents to start from copies of older Agents. This results in the deprecated `reparse_args` being used regularly in new code. Just trying to get things up to date.

## How Has This Been Tested?

I've tested that style of change on other Agents, though really I should test these Agents too. I will open this as draft until those are done (likely early next week.)

## Types of changes

## Checklist:

### DIFF
--- a/agents/fake_data/Dockerfile
+++ b/agents/fake_data/Dockerfile
@@ -1,5 +1,5 @@
-# OCS Registry Agent
-# ocs Agent container for running the registry.
+# OCS Fake Data Agent
+# ocs Agent container for generating random data
 
 # Use ocs base image
 FROM ocs:latest
@@ -7,6 +7,7 @@ FROM ocs:latest
 # Set the working directory to registry directory
 WORKDIR /app/ocs/agents/fake_data/
 
+# Copy this agent into the WORKDIR
 COPY . .
 
 # Run registry on container startup
--- a/agents/influxdb_publisher/influxdb_publisher.py
+++ b/agents/influxdb_publisher/influxdb_publisher.py
@@ -140,13 +140,8 @@ if __name__ == '__main__':
     # Start logging
     txaio.start_logging(level=environ.get("LOGLEVEL", "info"))
 
-    parser = site_config.add_arguments()
-
-    parser = make_parser(parser)
-
-    args = parser.parse_args()
-
-    site_config.reparse_args(args, 'InfluxDBAgent')
+    parser = make_parser()
+    args = site_config.parse_args(agent_class='InfluxDBAgent', parser=parser)
 
     agent, runner = ocs_agent.init_site_agent(args)
 
--- a/docs/developer/agents.rst
+++ b/docs/developer/agents.rst
@@ -126,16 +126,13 @@ A simple example of this process can be found in the HWP Simulator Agent:
   if __name__ == '__main__':
 
     # Create an argument parser
-    parser = site_config.add_arguments()
+    parser = argparse.ArgumentParser()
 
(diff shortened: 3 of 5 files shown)


mhasself:
The example code that you fixed up is a little problematic still because it doesn't add any agent-specific command-line args.  (--port would be a good one for demo.)

The documentation in https://ocs.readthedocs.io/en/develop/developer/site_config.html#agent-site-related-command-line-parameters is pretty good, but it is not linked from the Developers:Agents section... it should be.

The now-deprecated sequence is:
```
    site_parser = site_config.add_arguments()        # Get OCS base argparser
    parser = make_parser(site_parser)                # Add some Agent-specific args
    args = parser.parse_args()                       # Run the parser magic
    site_config.reparse_args(args, 'LabJackAgent')   # Fill in missing stuff with OCS rules
```
and that should be pretty close to the same thing as:
```
parser = make_parser()
args = site_config.parse_args(agent_class='LabJackAgent', parser=parser)
```
(Assuming of course that make_parser() can indeed make its own parser; otherwise pass in a new ArgumentParser().)

Were there SOCS agents using it in a weird way?

site_config.parse_args() seems suitable for agents but it is a little problematic for non-agents, such as clients and ocsbow... I think I have a fix for ocsbow at least; let me know if I should add it here or submit separately.

BrianJKoopman:
> The example code that you fixed up is a little problematic still because it doesn't add any agent-specific command-line args. (--port would be a good one for demo.)

Right, good point. Since the example in the docs uses the long unused HWP Simulator Agent that doesn't add any arguments I thought I'd swap it out for a perhaps more familiar example, the FakeDataAgent. I've done that, and replaced instances of full file insertions with `.. include::` directives so things stay up to date. That's done in 6979f6f.

> The documentation in https://ocs.readthedocs.io/en/develop/developer/site_config.html#agent-site-related-command-line-parameters is pretty good, but it is not linked from the Developers:Agents section... it should be.

Ah, great, didn't realize that was there, it's now linked. Updated in 347d305.

> Were there SOCS agents using it in a weird way?

I don't think so, it was just me trying to update it without knowing the full replacement, leaving behind an `site_config.add_arguments()` to make the initial parser, which caused the conflicts.

> site_config.parse_args() seems suitable for agents but it is a little problematic for non-agents, such as clients and ocsbow... I think I have a fix for ocsbow at least; let me know if I should add it here or submit separately.

I'd say a fix for ocsbow could be separate.

mhasself (approved):
Sorry to sleep on this, if you were waiting for me.  This is a draft PR but I guess it's ready to go?  This looks good to me.

BrianJKoopman:
> Sorry to sleep on this, if you were waiting for me. This is a draft PR but I guess it's ready to go? This looks good to me.

Thanks. I think I stalled out here when trying to write some tests for the behavior I thought I was trying to avoid. Going to go ahead with the merge now though after a rebase.

PR #202, opened 2021-04-20
Clarify the permissions structure for ocs-util

BrianJKoopman:
## Description

Clarify the permissions structure for ocs-util. Since this requires creating the ocs user, we move that information from the Aggregator Agent page to the more general (and visible) location in the installation instructions.

We also set and expose the same port as is recommended for use in the ocs-util instructions. This should help users who might run the container directly, rather than through docker-compose, else they might run into the situation where they try to access Jupyter via the recommended port 8880, but the service is running on the default 8888.

## Motivation and Context

Clarification prompted by user inquiry.

## How Has This Been Tested?

I've setup the ocs-util container with the recommended configuration and can write new Jupyter notebooks to disk.

## Types of changes

- [x] Documentation fix (updates to documentation)

## Checklist:


jlashner (approved):
Thanks Brian! This looks good.

PR #206, opened 2021-07-02
Check if args are passed to MatchedClient

### LINKED ISSUE #176: MatchedClient(instance_id) causes argparse to run on sys.argv
If you do something simple like
```
  c = MatchedClient('acu1')
```
then sys.argv will get run through the site_config parser.  This causes trouble when you want to use MatchedClient in anything where sys.argv is non-trivial, such as in jupyter or a script with its own command line args.  The work-around is to pass args=[] to this constructor.  But it would be better if by default there was no inspection of sys.argv.

The MatchedClient code should be changed as follows:  If args=None is passed in to the constructor, that should result in args=[] being passed through to site_config.  (If a user really wants to pass sys.argv through, they can do that explicitly in MatchedClient instantiation.)

BrianJKoopman:
## Description

Add a check when a MatchedClient is instantiated to see if `args` is passed in the `**kwargs`. If it's not, pass `args=[]` when getting the client, preventing inspection of `sys.argv`, which causes the site_config parser some problems.

## Motivation and Context

fixes #176

## How Has This Been Tested?

I wrote a test that would fail with the `'example_client.py: error: unrecognized arguments: test'` error we'd see if, for example, you tried to write a client that took additional command line arguments. It mocks up a lot, but worked for this test, and now passes.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/matched_client.py
+++ b/ocs/matched_client.py
@@ -92,6 +92,9 @@ class MatchedClient:
         For additional kwargs see site_config.get_control_client.
 
         """
+        if kwargs.get('args') is None:
+            kwargs['args'] = []
+
         self._client = site_config.get_control_client(instance_id, **kwargs)
         self.instance_id = instance_id
 
--- a/ocs/matched_client.py
+++ b/ocs/matched_client.py
@@ -79,17 +79,17 @@ class MatchedClient:
     """
 
     def __init__(self, instance_id, **kwargs):
-        """MatchedClient __init__ function.
-
+        """
         Args:
             instance_id (str): Instance id for agent to run
             args (list or args object, optional):
                 Takes in the parser arguments for the client.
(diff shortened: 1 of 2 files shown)


mhasself (approved):
Thanks!  The ```__init__``` docstring needs to be updated, then good to go.  Although this is an interface change, I doubt anyone will be bothered by this.

Didn't know about ```@patch``` ... will start using it!

BrianJKoopman:
> Thanks! The `__init__` docstring needs to be updated, then good to go. Although this is an interface change, I doubt anyone will be bothered by this.

Ah, great point. Updated with some other small formatting adjustments. Will merge when checks complete.

> Didn't know about `@patch` ... will start using it!

New to me too, definitely very useful!

PR #167, opened 2020-09-11
Requirements prune

BrianJKoopman:
## Description

This splits out the requirements so users can install a minimal amount of packages just required for running ocs clients, while optionally being able to install all packages needed for development, etc.

Left as a draft for now, the docs still need updating to explain the structure. If people have feedback for now though, it's definitely welcome.

Other notes:
* Removed ipywidgets requirement by also removing old ocs_widgets.py module.
* Removed msgpack, as it didn't seem to be used in anything.

## Motivation and Context

Closes #92 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] Unless I am preparing a release, I have opened this PR onto the `develop` branch.

PR #208, opened 2021-07-16
Deal gracefully with crossbar not being found

cnweaver:
Crossbar was not included in requirements.txt, which leads to some obscure errors from a fresh install on a clean system. 

## Description
This patch adds crossbar to requirements.txt so that users should generally not see this problem at all, and adds error checking to detect when crossbar cannot be found at runtime or the user has manually specified an invalid path to it. 

## Motivation and Context
This change prevents or renders more easy to understand problems with crossbar not being found. 

## How Has This Been Tested?
A new unit test is included to verify the behavior of these changes. 
Otherwise, the behavior can be verified by testing installation on any system which does not already have crossbar installed. 

## Types of changes

## Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,6 +50,9 @@ There are also several Python package dependencies, which are listed in the
 
 .. _requirements.txt: requirements.txt
 
+Additionally ``crossbar`` is a required dependency which is not automatically
+installed. It can be installed with pip or run via Docker.
+
 Installation
 ------------
 Clone this repository and install using pip::
--- a/docs/user/dependencies.rst
+++ b/docs/user/dependencies.rst
@@ -32,7 +32,7 @@ Other Dependencies
 * `Docker Compose`_ - CLI tool for running multi-container Docker
   applications.
 
-Crossbar ay be installed via pip if it will be used without Docker; if it
+Crossbar may be installed via pip if it will be used without Docker; if it
 will be used with Docker it does not need to be separately installed.
 
 Operating System
--- a/docs/user/dependencies.rst
+++ b/docs/user/dependencies.rst
@@ -20,17 +20,20 @@ Python Dependencies
 All python dependencies for OCS are listed in (and automatically installed via)
 the ``requirements.txt`` file. Some of the dependencies are:
 
-* `crossbar.io`_ - An implementation of a WAMP router.
 * `Autobahn`_ - Provides open-source implementations of WAMP.
 * `twisted`_ - Used with Autobahn for networking.
 
 Other Dependencies
 ``````````````````
 
+* `crossbar.io`_ - An implementation of a WAMP router.
 * Docker_ - Containerization software used for deploying several SO written
   packages.
 * `Docker Compose`_ - CLI tool for running multi-container Docker
   applications.
+
+Crossbar ay be installed via pip if it will be used without Docker; if it
+will be used with Docker it does not need to be separately installed.
 
 Operating System
 ````````````````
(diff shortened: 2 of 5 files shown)


BrianJKoopman:
Hi Chris, thanks for the PR. Can you provide some more details on the obscure errors you ran into? If I had to guess, they're likely related to `ocsbow` trying to startup and interact with the `crossbar` server?

We used to have `crossbar` in our requirements file. Typically now the `crossbar` server will be left running in a Docker container on a system and not started/stopped by `ocsbow`. Back in Sept 2020 we were running into the `crossbar` package pinning dependencies that would then had further consequences (see #166 and #173), so we dropped it.

`crossbar` itself pulls in a ton of dependencies (as pointed out in #92). I, at one point, started to split out the various requirements to separate files for different purposes (i.e. if you only need to run an ocs client), but never really wrapped it up (that's the draft sitting in #167).

cnweaver:
Indeed, the failure occurs when `ocsbow` tries to start `crossbar`, but the cause isn't very clear at first glance:
```$ make up
ocsbow --site-file default.yaml up
Trying to start crossbar...
  File "/home/cweaver/daq/bin/ocsbow", line 4, in <module>
    __import__('pkg_resources').run_script('ocs==0.8.0+8.gedccf36', 'ocsbow')
  File "/home/cweaver/daq/lib/python3.8/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/cweaver/daq/lib/python3.8/site-packages/pkg_resources/__init__.py", line 1462, in run_script
    exec(code, namespace, namespace)
  File "/home/cweaver/daq/lib/python3.8/site-packages/ocs-0.8.0+8.gedccf36-py3.8.egg/EGG-INFO/scripts/ocsbow", line 7, in <module>
    ocsbow.main()
  File "/home/cweaver/daq/lib/python3.8/site-packages/ocs-0.8.0+8.gedccf36-py3.8.egg/ocs/ocsbow.py", line 518, in main
    hm.crossbar_action('start')
  File "/home/cweaver/daq/lib/python3.8/site-packages/ocs-0.8.0+8.gedccf36-py3.8.egg/ocs/ocsbow.py", line 244, in crossbar_action
    pid = os.spawnv(flags, cmd[0], cmd)
  File "/usr/lib/python3.8/os.py", line 879, in spawnv
    return _spawnvef(mode, file, args, None, execv)
  File "/usr/lib/python3.8/os.py", line 846, in _spawnvef
    raise ValueError('argv first element cannot be empty')
ValueError: argv first element cannot be empty
make: *** [Makefile:15: up] Error 1
$ which crossbar || echo "not found" 1>&2
not found```

If you wanted to treat installation of crossbar separately I think that could be fine as long as it's made clear in the docs that the user needs to, as a deliberate step, follow one of the options to get it installed, and I think having this improved error message when the user misses that will help. Personally, I prefer just installing crossbar into the virtual environment in which I'm working, since Docker is a much heavier dependency, and often requires elevated privileges. 

If you prefer, I can remove the addition to `requirements.txt` from this PR, and instead update the docs to make it more clear that crossbar is _not_ installed with pip (as it is currently implied to be).

BrianJKoopman:
> If you wanted to treat installation of crossbar separately I think that could be fine as long as it's made clear in the docs that the user needs to, as a deliberate step, follow one of the options to get it installed, and I think having this improved error message when the user misses that will help. Personally, I prefer just installing crossbar into the virtual environment in which I'm working, since Docker is a much heavier dependency, and often requires elevated privileges.

I agree that the improved error message would help.

> If you prefer, I can remove the addition to `requirements.txt` from this PR, and instead update the docs to make it more clear that crossbar is _not_ installed with pip (as it is currently implied to be).

That sounds good to me. Thanks!

cnweaver:
This is now done.

BrianJKoopman (approved):
Looks good, thanks for the updates and the patch!

PR #212, opened 2021-08-02
Reorganize Dockerfile to leverage build cache

BrianJKoopman:
## Description

Add an explicit copy of the requirements file and call the pip install on it
earlier so that on modification of the ocs package we don't have to download
all dependencies again. This'll improve turn around time when iterating on
building OCS Docker images locally for testing during development.

## Motivation and Context

I was tired of waiting for PyPI downloads on rebuilds. Official builds will still need to download all packages fresh, as there is no cached image in that environment. So this will only really benefit local builds.

## How Has This Been Tested?

I've built the ocs images locally and run a local quickstart setup like in the docs.

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,11 +22,15 @@ RUN apt-get update && apt-get install -y python3 \
 # Install init system
 RUN pip3 install dumb-init
 
+# Copy in and install requirements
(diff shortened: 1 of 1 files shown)


BrianJKoopman:
Merging after a few of us discussed briefly on the phone.

PR #218, opened 2021-08-17
ocs-client-cli command-line tool

mhasself:
## Description

Creates script ocs-client-cli, to help quickly start an interactive python/ipython session with an open MatchedClient.  Also you can "scan" for Agents (and list their op's statuses).  Added to docs.

## Motivation and Context

This helps people avoid the boilerplate needed when all you want is a MatchedClient to debug / inspect a system.  When controlling special devices folks probably have Jupyter notebooks or longer client scripts they can run.  But there is a place for a general tool that can quickly connect to any Agent.  It's frankly pretty useful to be able to list active Agents from the command line.

Going forward we might want to add additional debugging capabilities, such as dumping Operation session.data to the terminal.

## How Has This Been Tested?

I've tested this on the miniobs, with both the python and ipython interpreters.

## Types of changes

## Checklist:


BrianJKoopman (changes requested):
This does look useful! I ran into one error that I think would be good to account for. Then I have a question (or maybe found a bug?) related to the `--details` behavior. Details in comments below.

PR #225, opened 2021-09-03
docs: add sphinx extension that auto-links to source code

mhasself:
As discussed.  This is a good thing.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,6 +42,7 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.napoleon',
     'sphinx.ext.todo',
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):

jlashner:
Does this automatically link to code from autodoc? Or is there something we need to add in the rst?

Just checked the ocs read the docs and can't find the links, but I don't really know where to look.

BrianJKoopman:
Next to autoclass or automethod class/method names there's a green [source]. That links not to the github repo, but to a sphinx page that shows the source file. For instance on the API page:
image

jlashner:
Ah got it, was looking at the "latest" build and not develop

PR #228, opened 2021-09-21
Only perform check on param if value is given

BrianJKoopman:
## Description

If a value isn't given, the check likely isn't valid. This will cause problems
if the decorated method is imported, for instance in any testing, as the value
will be None on import, which will fail simple checks like x < 100.

Example traceback when this happens:
```
tests/agents/test_ls372_agent.py:3: in <module>                                      
    from LS372_agent import LS372_Agent                                              
agents/lakeshore372/LS372_agent.py:70: in <module>                                   
    class LS372_Agent:                                                               
agents/lakeshore372/LS372_agent.py:585: in LS372_Agent                               
    ???                                                                              
../ocs/ocs/ocs_agent.py:1224: in param                                               
    ParamHandler({}).get(key, default=None, **kwargs)                                
../ocs/ocs/ocs_agent.py:1168: in get                                                 
    if not check(value):                                                             
agents/lakeshore372/LS372_agent.py:585: in <lambda>                                  
    ???                                                                              
E   TypeError: '<' not supported between instances of 'NoneType' and 'int'   
```

## Motivation and Context

No open issue for this bug, found when developing tests for an socs Agent.

## How Has This Been Tested?

The import in question above passes with this change.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/ocs_agent.py
+++ b/ocs/ocs_agent.py
@@ -1164,9 +1164,9 @@ class ParamHandler:
             if choices is not None:
                 if value not in choices:
                     raise ParamError(f"Param '{key}'={value} is not in allowed set ({choices})")
-        if check is not None:
-            if not check(value):
-                raise ParamError(f"Param '{key}' failed validity check (see docs?).")
+            if check is not None:
+                if not check(value):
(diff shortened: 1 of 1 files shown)


mhasself (approved):
Thanks!

PR #229, opened 2021-10-08
Update feed message data check to handle bools

BrianJKoopman:
## Description

If a `bool` is published to an OCS Feed it'll pass the data verification method that checks that data is of type `(float, int, str)` since `bool` is a subclass of `int`. Downstream this'll cause errors in the Aggregator (the Aggregator casts to a G3 type, sees it can't handle this data and drops it).

This patch will raise a `TypeError` on the Agent end if a `bool` is published.

## Motivation and Context

This issue was identified by @sanahabhimani on the system at Yale when a `bool` was making it to the Aggregator Agent. The Aggregator then continuously produced the error:
```
2021-10-05T20:00:02+0000 Error received when casting timestream! g3_cast does not support type <class 'bool'>. Type mustbe one of dict_keys([<class 'str'>, <class 'int'>, <class 'float'>])
```

## How Has This Been Tested?

Two tests have been added, asserting that the `TypeError` is raised when `bool`'s are passed either as a single sample, or in a list.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/ocs_feed.py
+++ b/ocs/ocs_feed.py
@@ -253,9 +253,11 @@ class Feed:
         """
         valid_types = (float, int, str)
 
+        # separate bool checks since bool is a subclass of int
         # multi-sample check
         if isinstance(value, list):
-            if not all(isinstance(x, valid_types) for x in value):
+            if (any(isinstance(x, bool) for x in value)
+                    or not all(isinstance(x, valid_types) for x in value)):
                 type_set = set([type(x) for x in value])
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
> I did not know that isinstance(a_bool, int)!

Yeah, I'll admit that was news to me too. Interesting Python history lesson -- `bool` wasn't added until version 2.3 in 2002 ([PEP 285](https://www.python.org/dev/peps/pep-0285/) -- which discusses the inheritance.)

PR #242, opened 2021-11-16
tests: test_params for the @params decorator system

mhasself:
Neglected to include this with the original param-helper PR!


BrianJKoopman (commented):
This is great, thanks Matthew!

A couple of comments:
- I want to point out that we're mixing test frameworks here. Our other tests are written using pytest, while these use Python's unittest. There's a page in the [pytest docs on unittest](https://docs.pytest.org/en/6.2.x/unittest.html). pytest supports running unittest based tests, but there's some notes on things that aren't supported. I think this is fine, I just wanted to point it out for future test development in case we run into any of those cases. I'd rather have mixed test frameworks than no tests at all.
- Organizationally I think this test class should go in `test_ocs_agent.py`, feel free to disagree.
- There's one missing test case, and that is of a param that can't be cast properly: https://coveralls.io/builds/44317374/source?filename=ocs%2Focs_agent.py#L1156

mhasself:
> This is great, thanks Matthew!
> 
> A couple of comments:
> 
> ...

Great points.  I never really clocked the difference between unittest and py.test!  should all be fixed in latest commits.

BrianJKoopman (approved):
Perfect, thanks!

PR #243, opened 2021-11-19
Build and Deploy to PyPI

BrianJKoopman:
## Description

This publishes OCS to PyPI. The first version of which is actually [already up](https://github.com/simonsobs/ocs/pull/243#issue-1058833674), but is currently somewhat broken (it doesn't pull in dependencies on install, so a `pip install -r requirements.txt` is still needed. This is fixed in this PR.)

Detailed breakdown of changes:
* Setup to build wheels and publish them to PyPI. The changes relevant to setting up for the build are in `pyproject.toml` and `setup.py`. 
* Update the README so that reading it on PyPI makes sense.
  * Relative links are now absolute so that they resolve from the PyPI page.
  * Update target Python version.
  * List dependencies that aren't pulled in automatically on pip install.
* Organize and prune some requirements. There's some overlap here in work that was started in #167.
* Add wheel build, test, and publish job to deploy workflow.
  * Also use more of the features of GH Actions, like 'needs', splitting up the jobs a bit (which I wanted to do originally, but must have missed the existence of 'needs').

## Motivation and Context

Work is ongoing to make so3g pip installable from PyPI. I was motivated by that work to get ocs on PyPI as well. This may have the added side effect of motivating more timely releases (which we're really overdue for at this point.)

## How Has This Been Tested?

I've built the wheel and installed locally. Also did that first publication to PyPI and tested that (and found it broken and have fixed here.) The wheel building and testing pipeline was tested on tag pushes before being merged into the deploy workflow, so I'm fairly confident that'll work when we make our next release.
* TestPyPI version: https://test.pypi.org/project/ocs/

## Types of changes

## Checklist:


mhasself (approved):
Wow, so cool!

The only odd thing is why there are any TLA available on PyPI ...

PR #256, opened 2022-02-03
Use pip installed so3g in GH Action workflows

BrianJKoopman:
## Description

Now that so3g is easily pip installable this allows us to do several things:
- Install so3g within the GH Actions workflows directly, meaning we can drop the use of docker containers for running tests that depend on spt3g/so3g.
- Remove the spt3g marker on our tests used to select only those test which do or do not depend on spt3g/so3g. We now just insist that so3g is installed for testing by adding it to the `requirements/testing.txt` file. Associated with this we no longer check for successful imports as a dependency for running those tests, meaning we can also drop the `pytest-dependency` module as a requirement.

I've also listed so3g as an optional requirement via `extras_install` in `setup.py`.

## Motivation and Context

Take advantage of the fact that so3g is now on PyPI and easily installable.

## How Has This Been Tested?

I've tested the test locally, and they will run in this PR/have run on these commits. The deploy develop and deploy workflow changes are currently untested, though mimic those changes made in the pytest workflow.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,27 +24,19 @@ jobs:
     - name: Install ocs
       run: |
         pip3 install -r requirements.txt
-        pip3 install -e .
+        pip3 install -e .[so3g]
 
     # Test (already been run by pytest workflow, but they don't take long...)
     # Unit Tests
     - name: Run unit tests
       working-directory: ./tests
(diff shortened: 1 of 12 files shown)


mhasself (approved):
Neat!

PR #276, opened 2022-07-27
docs: Address quickstart guide issues

### LINKED ISSUE #119: Set wamp_server and wamp_http for individual hosts

### LINKED ISSUE #210: Connecting to crossbar in various host/container configurations can be confusing
There's a subtle point of confusion that often comes up with configuration of the crossbar address in ocs-site-config files, especially when it comes to defaults within docker containers, docker's internal networking, and containers that run in "host" network mode.

This should be clarified in the docs, but perhaps also we should consider a way to improve this connection configuration.

### LINKED ISSUE #238: Make Quickstart more of a "Minimal Working System" (or add an example of that?)
The quickstart was great for getting an example of how stuff looks when it's running (ie. showing off grafana with fake data) but it has no saving / persistence across restarts.

Going that one more step to "just follow these instructions to get a minimal working / saving system" would be very nice and helpful. I really wanted to just get someone setup putting together an agent/client pair outside the LATRt system and the jump of "now go figure out how to make everything save" is a lot. All the systems I see to copy from have many stacked docker-compose files + many hosts so are a much more complicated reach.

BrianJKoopman:
## Description

This fixes several user reported issues in the quickstart guide, the big one being persistent storage of grafana/influxdb data. Other items on my list were:

- Add `sudo` to user creation steps.
- Add `-E` to sudo usage when running `docker-compose up -d`.
- Fix link to Docker installation docs.
- Remove mention of `setup.py` during installation.
- Point out Agent Reference pages when discussing the site config file.
- Persist influxdb and grafana storage in quickstart example. (Also added to the longer example under "System Configuration".)
- Warn users more clearly when removal of data volumes would cause data loss.
- Update the SCF/dockerfiles with a better recommendation for how to configure `wamp_server` location with respect to whether we're on the host system or in a docker container. (This is also a big one, and I hope will lead to less confusion about this, as it's perhaps the most common misconfiguration on new systems.)

## Motivation and Context

Resolves #238.
Resolves #210.
Resolves #119.

## How Has This Been Tested?

I ran through these changes in a clean VM, though I'm maybe not the best test...

## Types of changes

## Checklist:


BrianJKoopman:
Fixed some small warnings showing up on builds, and resolved some conflicts that arose after the ocs-web removal. Will merge when checks finish.

PR #279, opened 2022-07-27
Rewrite Agent developer guide

BrianJKoopman:
## Description, Motivation, and Context

This is a major rewrite of the Agent developer guide section. We've received feedback from users saying that the Fake Data Agent was a complicated example to use for this guide. To resolve this we introduce the "Barebones Agent", which is now the focus of the guide.

The new "Writing an Agent" guide starts from a completely empty Agent class that does nothing (i.e. only connects to the crossbar server, doesn't have any Tasks or Processes) and builds up from there. Each new page introduces a feature, and by the end we have a full example of an Agent that as one Task and one Process. 

I think this should help users, especially given one of the most common pieces of feedback was that it was hard to tell what was safe to strip out of the Fake Data Agent example.

I also took the opportunity to split the rest of the Agents page into several separate pages, what's left are pages for:
- Operation Parameters (the `@params` decorator)
- TimeoutLock
- Logging
- Documentation

These pages are meant to be more just a general reference, especially for those who are familiar already with writing Agents and just need to look something up. TimeoutLock is heavily covered now in the new guide, so maybe that could be removed/largely integrated into the guide, but I've left it for now.

## How Has This Been Tested?

I've run through the guide in a clean VM, after following the updated quick start guide in #276. My hope is to get someone to read through it and find any inconsistencies/places for improvement. Suggestions welcome!

## Checklist:


BrianJKoopman:
I'll likely try to merge this after our Wednesday call, any feedback before then is appreciated.

mhasself (approved):
I didn't read this fully ... but I scanned a lot of it and it looks awesome.  Minor comments.

PR #285, opened 2022-09-14
HostManager supports new plugin system

mhasself:
Note it still supports the old system, too.

I think it will be possible to limp along with both systems for a bit, which might ease transition.

Note I also enabled `python -m ocs.agent_cli`, which is a useful way to run it on a conda system without actually entering the conda env.


BrianJKoopman (approved):
Thanks for the updates! I agree about leaving the old system in place for a bit.

PR #288, opened 2022-10-19
Setup pre-commit and run on all files

BrianJKoopman:
## Description

This setups up pre-commit with a matching config to that used on [socs](https://github.com/simonsobs/socs/blob/develop/.pre-commit-config.yaml).

There's still one flake8 warning in ocsbow that maybe I'll ask @mhasself to take a look at:
```
ocs/ocsbow.py:830:67: F821 undefined name 'host_name'
```

Another warning that I fixed but was still wondering about related to `log_publish()` in `ocs.ocs_agent`. It looks like this maybe used to publish logs to a feed? Is this still doing anything? Formatting?

## Motivation and Context

It's been really useful on socs, and so I thought I'd set it up here as well.

## How Has This Been Tested?

Run pre-commit locally on all files. All tests still pass after making the included changes.

## Types of changes

## Checklist:


mhasself:
W.r.t. host_name I am 99.5% sure this is the right fix:
```
-                            inst['instance-id'], len(active_hms), host_name))
+                            inst['instance-id'], len(active_hms), host_data['host_name']))
```

mhasself:
I think it's ok to remove the log_publish method and the addObserver call.

I vaguely remember that logging to a feed caused problems when the feed couldn't be written to, because that raised an error, which got logged, which tried to write to the feed ...

PR #296, opened 2022-11-02
Adds wait-time arg to registry

jlashner:
Increases default wait time of registry and adds corresponding command line arg.

## Description
Using Matthews script to scan hk files I noticed that the registry was taking up a significant amount of disk space on the SAT1, since the number of fields increases with the number of other agents running on the system. This PR reduces the sleep time in the main loop from 1 sec to 30 sec to mitigate this, and adds the `wait-time` argument that allows users to set it in the sys-config.

## Motivation and Context

Disk space / grafana issues

## How Has This Been Tested?
Tested on my local desktop to make sure it runs.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/agents/registry/agent.py
+++ b/ocs/agents/registry/agent.py
@@ -127,7 +127,7 @@ class Registry:
         msg = {'block_name': addr,
                'timestamp': time.time(),
                'data': {}}
-        self.log.info(addr)
+        self.log.debug(addr)
         for op_name, op_code in reg_agent.op_codes.items():
             field = f'{addr}_{op_name}'
             field = field.replace('.', '_')
--- a/ocs/agents/registry/agent.py
+++ b/ocs/agents/registry/agent.py
@@ -4,6 +4,7 @@ import time
 from twisted.internet.defer import inlineCallbacks
 from autobahn.twisted.util import sleep as dsleep
 from ocs.ocs_feed import Feed
+import argparse
 
 
 class RegisteredAgent:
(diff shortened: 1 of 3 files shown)


BrianJKoopman (changes requested):
This'll definitely reduce the data volume. With long wait times, even the 30 seconds you're using here, I'm worried about missing state changes in the op codes the occur within the wait-time window. What do you think about publishing only on detected op code state changes, then occasionally (whatever the wait-time is set for) publishing everything, even if state hasn't changed?

I know this is time sensitive, so my proposal on the short term is to have the wait-time default to the old value of 1 second. This'll not change behavior for anyone who updates, but still allow you to reduce the data rate on the SAT1 setup. If that's alright with you, I'm fine merging with that one change.

jlashner:
I think having data published regularly is useful for alerts, so I"m hesitant to only publish data on operation change. This recent PR makes it so ops are published whenever an operation is changed, in addition to once every `wait-time` seconds which defaults to 30. Let me know if that works for you!

ocs/agents/registry/agent.py
```diff
@@ -115,7 +115,33 @@ def _register_heartbeat(self, _data):
...
         self.registered_agents[addr].refresh(op_codes=op_codes)
+        if publish:
+            self._publish_agent_ops(reg_agent)
+
+    def _publish_agent_ops(self, reg_agent):
+        addr = reg_agent.agent_address
+        msg = {'block_name': addr,
+               'timestamp': time.time(),
+               'data': {}}
+        self.log.info(addr)
```
BrianJKoopman, line 130: Maybe more appropriately a debug log? If not maybe some more info should be included, like "heartbeat for {addr} published" or similar?

BrianJKoopman (commented):
Yeah, perfect, that's what I was trying to ask for. Just one small comment on the log message.

BrianJKoopman (approved):

PR #298, opened 2022-11-04
Update ocs-local-support so it can launch HostManager again

### LINKED ISSUE #295: Wrong host_master script path in v0.10.0
I needed to re-install OCS recently, and updated to the latest release, v0.10.0 in the process. I was rather confused to get the following output when I attempted to start the system:

	$ ocs-local-support start
	
	Trying to launch hostmanager agent...
	  File "/home/cweaver/ocs/bin/ocs-local-support", line 8, in <module>
	    sys.exit(main_local())
	  File "/home/cweaver/ocs/lib/python3.8/site-packages/ocs/ocsbow.py", line 954, in main_local
	    raise OcsbowError('Failed to start manager process: %s' % message)
	ocs.ocsbow.OcsbowError: Failed to start manager process: Could not find host_manager.py in the agent_paths!

To verify my agent paths were sensible, I ran `find $VIRTUAL_ENV -name host_manager.py`, but it outputs nothing, as it appears this file has been renamed, although `ocs/ocsbow.py` still tries to refer to to is explicitly. Manually altering `ocs/ocsbow.py` to refer to `host_manager/agent.py` instead of `host_manager/host_manager.py` seems to have fixed it, allowing the system to start.

mhasself:
ocs-local-support needs to be able to launch the HostManager, but that bit of code wasn't set up to use the new agent plugin.

## Description

If a user runs "ocs-local-support start agent", the HostManager is launched in the background using python -m ocs.agents.host_manager.agent.  This is used instead of -m ocs-agent-cli, because additional command line args need to be passed through at run time.

## Motivation and Context

Resolves #295

This use case was missed in the recent upgrade to plugin system.  The recommended alternative to "ocs-local-support start agent" is to use systemd, and that was tested.

## How Has This Been Tested?

Tested on a local config that was also used to test other HostManager/plugin updates.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/ocsbow.py
+++ b/ocs/ocsbow.py
@@ -577,7 +577,7 @@ class HostManagerManager:
                 return True, 'Agent has exited and relinquished registrations.'
         return False, 'Agent did not die within %.1f seconds.' % timeout
 
-    def start(self, check=True, timeout=5., up=False, foreground=False):
+    def start(self, check=True, timeout=5., up=None, foreground=False):
         if self.instance_id is None:
             print('\nERROR: cannot "start" HostManager because '
                   'instance_id could not be determined.')
@@ -602,8 +602,10 @@ class HostManagerManager:
                '--site-file', self.site_config.site.source_file,
                '--site-host', host.name,
                '--working-dir', self.working_dir]
-        if up:
+        if up is True:
             cmd.extend(['--initial-state', 'up'])
+        if up is False:
+            cmd.extend(['--initial-state', 'down'])
         if not foreground:
             cmd.extend(['--quiet'])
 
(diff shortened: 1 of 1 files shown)


BrianJKoopman (commented):
This looks good, I'm able to start and stop host manger in my setup with:
```
$ OCS_CONFIG_DIR=$(pwd) ocs-local-support start agent --instance-id hm-mydaqhost1
$ OCS_CONFIG_DIR=$(pwd) ocs-local-support stop agent --instance-id hm-mydaqhost1
```

However, I'm finding that no matter what I put for `--initial-state` in my SCF it will send 'up' as the state. I think that's happening here:
https://github.com/simonsobs/ocs/blob/dbc645ff8935f32c7c3bea1059e8461e66eb8951/ocs/ocsbow.py#L963

Can we have this inspect the SCF value instead?

BrianJKoopman (approved):
Thanks for the update!

PR #301, opened 2022-11-08
Combine payloads before writing to influxdb

BrianJKoopman:
## Description

This PR combines the data in the queue before writing to the InfluxDB. 

Currently, the InfluxDB Publisher will issue separate `write_points()` calls for each item in the queue (data published on a feed). This results in increasingly more `write_points()` calls depending on how many feeds are being published to on the OCS network. `write_points()` is efficient on large payloads (we had previously set a "batch size" of 10,000 points). I don't think we'll be approaching this limit, but write now each payload is generally pretty small. 

This should improve write efficiency to the database a bit, though I'm not sure exactly how much.

## Motivation and Context

I was revisiting some of the structure of the publisher after recent reports of data lag in InfluxDB on several lab systems, most notably at UCSD.

## How Has This Been Tested?

Tested locally with 5 fake data agents writing to InfluxDB.

## Types of changes

## Checklist:

### DIFF
--- a/docs/agents/influxdb_publisher.rst
+++ b/docs/agents/influxdb_publisher.rst
@@ -61,6 +61,8 @@ online at all times. An example compose file would look like::
       restart: always
       ports:
         - "8086:8086"
+      environment:
+        - INFLUXDB_HTTP_LOG_ENABLED=false
       volumes:
         - /srv/influxdb:/var/lib/influxdb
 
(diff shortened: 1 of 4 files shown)


BrianJKoopman:
While perhaps not really related, I added lines to disable the HTTP logs in influxdb in [1ef573f](https://github.com/simonsobs/ocs/pull/301/commits/1ef573fa45e524c54385904f2ad4ddfe9e962e0f). We made this recommendation internally earlier this year. While this PR will reduce the amount of HTTP logs dramatically it'll still be logging ~1 request/second, so I think we should continue to disable them.

PR #304, opened 2022-11-11
Sanitize session data

### LINKED ISSUE #300: NaNs / infs in session.data cause trouble in ocs-web
NaNs/infs are not compliant JSON, but are passed through from Agents to ocs-web anyway.  This causes ocs-web to totally choke on the session.data ... which is difficult to avoid.

I don't think there's a way to catch this or change how JSON is parsed by autobahn.

Two possible solutions:
1. Convert NaNs/infs to None in ocs_agent.OpSession; this will produce compliant JSON.
2. Reject NaNs/infs entirely -- i.e. forbid them forcibly and raise an exception

I think solution (1) is slightly better, but it should be noted that some Python based clients may be affected by this.  First, one can not distinguish between NaN, inf, -inf.  I think that's ok ... usually we're just trying to flag invalid data, using NaN.

Another consequence of using "None" (which renders as "null" in JSON) is that lists received by Python clients will not as easily be convertible into arrays.  E.g. if the agent sends [1.0, 2.0, None], and the client tries to turn that into a numpy array:
```
>>> np.array([1.0, 2.0, None])
array([1.0, 2.0, None], dtype=object)
```
(that's bad -- it's of dtype=object).

But a work-around on the client side is to specify dtype:
```
>>> np.array([1.,2.,None], dtype=float)
array([ 1.,  2., nan])
```

I think this is an acceptable price to pay, to have ocs-web work and also allow NaNs.

mhasself:
Resolves #300 

## Description

- It's now ok to include numpy arrays / scalars in session.data -- they will be translated to simple lists of python scalars
- It's now ok to include NaN in session.data, it will be converted to None.
- It's now explicitly not ok to include +/- inf in session.data, an internal error will result.
- There are tests for the above.
- The expectations and behavior are documented.

This is a slight "break" in interface in that Inf/-Inf are now expressly forbidden, and NaN will render differently in clients that used to accept it (i.e. Python clients).  As discussed in the issue, this is deemed acceptable.

## How Has This Been Tested?

With tests.  Also with agent emulator in ocs-web.

## Types of changes

## Checklist:

### DIFF
--- a/docs/developer/session-data.rst
+++ b/docs/developer/session-data.rst
@@ -147,6 +147,9 @@ This block formats the latest values for each "channel" into a dictionary and
 stores it in ``session.data``.
 
 
+Structure and Content
+---------------------
+
 The structure of the ``data`` entry is not strictly defined, but
 please observe the following guidelines:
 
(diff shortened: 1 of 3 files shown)


mhasself:
Thanks @BrianJKoopman  -- I pulled in your fix, fixed the docs as you requested, and added one more test case.

BrianJKoopman (approved):
Awesome, looks good!

PR #312, opened 2022-12-15
Separate blocks for each registered operation

### LINKED ISSUE #311: OCS registry fails when block structure changes
This is likely caused by an agent being restarted / updated with additional operations. Registry needs to be resiliant to this. Maybe changing the structure so each operation instead of each agent has it's own block.

```
2022-12-14T20:39:36+0000 OCSAgent.onUserError(): "While firing <bound method Registry._register_heartbeat of <ocs.agents.registry.agent.Registry object at 0x7f4f02f3a4f0>> subscribed under 5302732399893542."
  File "/usr/local/lib/python3.8/dist-packages/autobahn/twisted/websocket.py", line 387, in _onMessage
    self.onMessage(payload, isBinary)
  File "/usr/local/lib/python3.8/dist-packages/autobahn/wamp/websocket.py", line 106, in onMessage
    self._session.onMessage(msg)
  File "/usr/local/lib/python3.8/dist-packages/autobahn/wamp/protocol.py", line 791, in onMessage
    future = txaio.as_future(handler.fn, *invoke_args, **invoke_kwargs)
  File "/usr/local/lib/python3.8/dist-packages/txaio/tx.py", line 369, in as_future
    return maybeDeferred(fun, *args, **kwargs)
--- <exception caught here> ---
  File "/usr/local/lib/python3.8/dist-packages/twisted/internet/defer.py", line 206, in maybeDeferred
    result = f(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/ocs/agents/registry/agent.py", line 122, in _register_heartbeat
    self._publish_agent_ops(reg_agent)
  File "/usr/local/lib/python3.8/dist-packages/ocs/agents/registry/agent.py", line 142, in _publish_agent_ops
    self.agent.publish_to_feed('agent_operations', msg)
  File "/usr/local/lib/python3.8/dist-packages/ocs/ocs_agent.py", line 540, in publish_to_feed
    self.feeds[feed_name].publish_message(message)
  File "/usr/local/lib/python3.8/dist-packages/ocs/ocs_feed.py", line 224, in publish_message
    b.append(message)
  File "/usr/local/lib/python3.8/dist-packages/ocs/ocs_feed.py", line 36, in append
    raise Exception("Block structure does not match: {}".format(self.name))
builtins.Exception: Block structure does not match: observatory.pysmurf-controller-c2s2
```

jlashner:
Changes registry block structure so that each operation has its own block.

## Description
It is possible for the operation list of an agent to change, if that agent is updated without restarting the registry. This will crash the registry and prevent it from publishing data since the block-structure of the feed depends on the available agents. This changes the registry so that each operation has its own block, allowing for the operation list to change without crashing the registry.

## Motivation and Context
Solves issues where updating agents without restarting the registry causes it to crash. 
Resolves #311.

## How Has This Been Tested?

Ran registry tests locally.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/agents/registry/agent.py
+++ b/ocs/agents/registry/agent.py
@@ -135,8 +135,8 @@ class Registry:
                 self.log.warn(f"Improper field name: {field}\n{e}")
                 continue
             msg = {'block_name': field,
-                'timestamp': time.time(),
-                'data': dict(field=op_code)}
+                   'timestamp': time.time(),
+                   'data': dict(field=op_code)}
             self.agent.publish_to_feed('agent_operations', msg)
 
     @ocs_agent.param('test_mode', default=False, type=bool)
--- a/ocs/agents/registry/agent.py
+++ b/ocs/agents/registry/agent.py
@@ -145,7 +145,7 @@ class Registry:
                 continue
             msg = {'block_name': field,
                    'timestamp': time.time(),
-                   'data': dict(field=op_code)}
+                   'data': {field: op_code}}
             self.agent.publish_to_feed('agent_operations', msg)
 
     @ocs_agent.param('test_mode', default=False, type=bool)
--- a/ocs/agents/registry/agent.py
+++ b/ocs/agents/registry/agent.py
@@ -123,9 +123,6 @@ class Registry:
 
     def _publish_agent_ops(self, reg_agent):
         addr = reg_agent.agent_address
-        msg = {'block_name': addr,
-               'timestamp': time.time(),
-               'data': {}}
         self.log.debug(addr)
         for op_name, op_code in reg_agent.op_codes.items():
             field = f'{addr}_{op_name}'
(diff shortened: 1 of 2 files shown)


ocs/agents/registry/agent.py
```diff
@@ -146,8 +143,9 @@ def _publish_agent_ops(self, reg_agent):
             except ValueError as e:
                 self.log.warn(f"Improper field name: {field}\n{e}")
                 continue
-            msg['data'][field] = op_code
-        if msg['data']:
+            msg = {'block_name': field,
+                   'timestamp': time.time(),
+                   'data': dict(field=op_code)}
```
BrianJKoopman, line 148: This use of `dict()` results in `'data': {'field': 3}` (if `op_code==3`, for instance). Down the line this changes the field name that should be 
```
'observatory.registry.feeds.agent_operations.observatory_aggregator_record'
```
into
```
'observatory.registry.feeds.agent_operations.field'
```
If you want to use the `dict()` constructor you need to pass a list of tuples, i.e. `'data': dict([(field, op_code)])`.

Or you can use `'data': {field: op_code}`, which would be my preference.

BrianJKoopman (changes requested):
Sorry this took so long for me to test. I was wondering if there were any implications past the registry in the system of splitting up the blocks and wanted to AB test opening files with and w/o the patch. In doing so I did notice a bug in the field name assignment.

ocs/agents/registry/agent.py
jlashner: This is a good catch, thanks! I'm totally fine with switching to the regular dict format.

BrianJKoopman (approved):
Thanks for the update. Tested one more time, things look good to me now!

PR #314, opened 2022-12-17
Test when registry block changes

BrianJKoopman:
## Description

This introduces a test that checks an exception is raised when a new operation is added to an Agent while the registry is running. This is currently a bug, but a fix would cause this test to fail. When fixed, the check for an exception should be removed.

I also added a small bit of documentation for some things that I needed to figure out while writing this.

## Motivation and Context

This tests the behavior of https://github.com/simonsobs/ocs/issues/311. Removing the check for an exception would be a good test of https://github.com/simonsobs/ocs/pull/312.

## How Has This Been Tested?

Ran locally.

## Types of changes

## Checklist:


jlashner (approved):
Hey Brian, thanks this looks great! I don't see anything that needs changing, but is the point of creating the `OCSAgent` object instead of mocking it so that the publish function will do all the feed-checking logic etc.?

BrianJKoopman:
> [I]s the point of creating the `OCSAgent` object instead of mocking it so that the publish function will do all the feed-checking logic etc.?

Yup, exactly! That change hasn't affected other test either, which is nice.

BrianJKoopman:
Just rebased, will merge in a moment.

PR #325, opened 2023-04-12
ocs-agent-cli renames itself in process list

mhasself:
Instead of showing up in top as "ocs-agent-cli", it shows up as "ocs-agent:{instance_id}".

This is a minor quality of life improvement ... sometimes you need to kill a process, or see who's using all the memory / CPU, and it's not obvious if all agents launched through ocs-agent-cli simply report themselves as "ocs-agent-cli".

### DIFF
--- a/ocs/agent_cli.py
+++ b/ocs/agent_cli.py
@@ -1,6 +1,7 @@
 import argparse
 import importlib
 import os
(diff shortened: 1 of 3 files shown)


BrianJKoopman (approved):
Great idea!

PR #327, opened 2023-05-10
Revist address_root != "observatory"

mhasself:
## Description

The `address_root` has always been set to `observatory` in OCS installations.  But this prefix is configurable in the SCF, and for the most part wasn't hard-coded anywhere in ocs (nor socs).

I have removed a small number of hard-coded occurrences of `observatory.`.  These were in the feed subscription wildcards the registry, aggregator and influxdbpublisher.

The address_root was intended to be user configurable; I consider this to be a bugfix.

Some additional changes along the way:
- I have purged references to the "registry_address" setting in SCF, which actually has been obsolete for a long time.
- InfluxDBPublisher agent is slightly modified, so it will exit on SIGINT (ctrl-c) while failing to find the influxdb server.
- ocs-local-support code is fixed to distinguish correctly between (a) crossbar is down (b) crossbar is up but realm and/or prefix are mismatched (c) crossbar is up and respon

## Motivation and Context

At the observatory we'll have a few very similar systems, e.g. sat1 and sat2, running as independent OCS instances.  In some situations (influxdb and grafana, primarily) we will have data from all systems available, and so the feeds can't have the exact same names.  That means we either need to carefully include "sat1" / "sat2" in every instance-id ... e.g. "observatory.sat1_acu", "observatory.sat2_acu"  OR take advantage of the configurable prefix and have the agent addresses be "sat1.acu", "sat2.acu".

## How Has This Been Tested?

I ran an OCS setup on a single host, with all the affected agents, plus a fake data agent.  Confirmed I could see the data appear in grafana and that data was showing up in aggregator output files.  All tests/ pass.

## Types of changes

## Checklist:

### DIFF
--- a/docs/user/site_config.rst
+++ b/docs/user/site_config.rst
@@ -120,6 +120,13 @@ and feed addresses on the crossbar network.  While "observatory" is
 the default, it can be changed as long as the crossbar configuration
 is also updated to permit operations on the `{address_root}.` uri.
 
+.. note::
+   The hub settings must match the crossbar configuration.  If you
+   change `wamp_realm` or `address_root`, especially, be sure to
+   update your crossbar configuration accordingly.  (If using the
+   ocs-crossbar docker image, this can be done through environment
+   variables in the docker-compose.yaml file.)
+
 Under `hosts` we have defined a three hosts, `host-1`, `host-1-docker`, and
 `host-2`. This configuration example shows a mix of Agents running directly on
 hosts and running within Docker containers.
--- a/docs/user/site_config.rst
+++ b/docs/user/site_config.rst
@@ -120,12 +120,12 @@ and feed addresses on the crossbar network.  While "observatory" is
 the default, it can be changed as long as the crossbar configuration
 is also updated to permit operations on the `{address_root}.` uri.
 
-.. note::
+.. warning::
    The hub settings must match the crossbar configuration.  If you
    change `wamp_realm` or `address_root`, especially, be sure to
    update your crossbar configuration accordingly.  (If using the
    ocs-crossbar docker image, this can be done through environment
-   variables in the docker-compose.yaml file.)
+   variables in the ``docker-compose.yaml`` file.)
 
 Under `hosts` we have defined a three hosts, `host-1`, `host-1-docker`, and
 `host-2`. This configuration example shows a mix of Agents running directly on
--- a/docs/user/site_config.rst
+++ b/docs/user/site_config.rst
@@ -35,7 +35,6 @@ instances (running two different classes of agent):
     wamp_http: http://10.10.10.3:8001/call
     wamp_realm: test_realm
     address_root: observatory
-    registry_agent: observatory.registry
 
   hosts:
 
--- a/docs/user/site_config.rst
+++ b/docs/user/site_config.rst
@@ -116,6 +116,11 @@ The `hub` section defines the connection parameters for the crossbar server.
 This entire section will likely remain unchanged, except for the
 ``wamp_server`` and ``wamp_http`` IP addresses.
 
+The `address_root` setting determines the leading token in all agent
+and feed addresses on the crossbar network.  While "observatory" is
+the default, it can be changed as long as the crossbar configuration
+is also updated to permit operations on the `{address_root}.` uri.
+
 Under `hosts` we have defined a three hosts, `host-1`, `host-1-docker`, and
 `host-2`. This configuration example shows a mix of Agents running directly on
 hosts and running within Docker containers.
--- a/ocs/agents/influxdb_publisher/drivers.py
+++ b/ocs/agents/influxdb_publisher/drivers.py
@@ -51,6 +51,9 @@ class Publisher:
             Protocol for writing data. Either 'line' or 'json'.
         gzip (bool, optional):
             compress influxdb requsts with gzip
+        operate_callback (callable, optional):
+            Function to call to see if failed connections should be
+            retried (to prevent a thread from locking).
 
     Attributes:
         host (str):
@@ -66,7 +69,8 @@ class Publisher:
 
     """
 
-    def __init__(self, host, database, incoming_data, port=8086, protocol='line', gzip=False):
+    def __init__(self, host, database, incoming_data, port=8086, protocol='line',
+                 gzip=False, operate_callback=None):
         self.host = host
         self.port = port
         self.db = database
@@ -88,6 +92,9 @@ class Publisher:
                 LOG.error("Connection error, attempting to reconnect to DB.")
                 self.client = InfluxDBClient(host=self.host, port=self.port, gzip=gzip)
                 time.sleep(1)
+            if operate_callback and not operate_callback():
+                break
+
         db_names = [x['name'] for x in db_list]
 
         if self.db not in db_names:
--- a/docker/crossbar/Dockerfile
+++ b/docker/crossbar/Dockerfile
@@ -8,8 +8,11 @@ FROM crossbario/crossbar:cpy3-20.8.1
 # Run as root to put config in place and chown
 USER root
 
-# Copy in config and requirements files
-COPY config.json /ocs/.crossbar/config.json
+# Copy in config template and wrapper script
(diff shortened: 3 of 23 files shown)


ocs/agents/influxdb_publisher/drivers.py
```diff
@@ -88,6 +92,9 @@ def __init__(self, host, database, incoming_data, port=8086, protocol='line', gz
                 LOG.error("Connection error, attempting to reconnect to DB.")
                 self.client = InfluxDBClient(host=self.host, port=self.port, gzip=gzip)
                 time.sleep(1)
+            if operate_callback and not operate_callback():
+                break
```
BrianJKoopman, line 96: This is neat, but I don't quite follow how this allows a ctrl+c interruption. Does the ctrl+c result in a `stop` call, which makes `self.aggregate == False` so `operate_callback()` is `False`, then we hit the `break`?

docs/user/site_config.rst
```diff
@@ -116,6 +115,11 @@ The `hub` section defines the connection parameters for the crossbar server.
 This entire section will likely remain unchanged, except for the
 ``wamp_server`` and ``wamp_http`` IP addresses.
 
+The `address_root` setting determines the leading token in all agent
+and feed addresses on the crossbar network.  While "observatory" is
+the default, it can be changed as long as the crossbar configuration
+is also updated to permit operations on the `{address_root}.` uri.
```
BrianJKoopman, line 121: I suppose my only concern now is users now having to configure crossbar, which we were avoiding previously. I think it would be easy to miss this requirement that the crossbar URI needs to change, especially since users haven't been dealing with crossbar configs previously.

`ocsbow status` will detect a potential mismatch, but just looking at agent log output it looks like a simple crossbar server not available and attempted reconnect:
```
2023-05-12T21:12:08+0000 session joined: {'authextra': {'x_cb_node': 'efc3505b29a0-1',
               'x_cb_peer': 'tcp4:192.168.96.7:57996',
               'x_cb_pid': 17,
               'x_cb_worker': 'worker001'},
 'authid': 'HG7R-VKER-KUTG-FR4V-JHS7-M7RP',
 'authmethod': 'anonymous',
 'authprovider': 'static',
 'authrole': 'iocs_agent',
 'realm': 'test_realm',
 'resumable': False,
 'resume_token': None,
 'resumed': False,
 'serializer': 'msgpack.batched',
 'session': 7293750874092658,
 'transport': {'channel_framing': 'websocket',
               'channel_id': {},
               'channel_serializer': None,
               'channel_type': 'tcp',
               'http_cbtid': None,
               'http_headers_received': None,
               'http_headers_sent': None,
               'is_secure': False,
               'is_server': False,
               'own': None,
               'own_fd': -1,
               'own_pid': 7,
               'own_tid': 7,
               'peer': 'tcp4:192.168.96.2:8001',
               'peer_cert': None,
               'websocket_extensions_in_use': None,
               'websocket_protocol': None}}
2023-05-12T21:12:08+0000 Failed to register basic handlers @ test2.fake-data1; agent probably running already.
2023-05-12T21:12:08+0000 session left: CloseDetails(reason=<wamp.close.normal>, message='None')
2023-05-12T21:12:08+0000 transport disconnected
2023-05-12T21:12:08+0000 waiting for reconnect for 9.999993801116943 more seconds
2023-05-12T21:12:08+0000 Scheduling retry 1 to connect <twisted.internet.endpoints.TCP4ClientEndpoint object at 0x7fc9ac4b2f10> in 2.247658769532999 seconds.
2023-05-12T21:12:08+0000 waiting for reconnect for 1.9931435585021973 more seconds
2023-05-12T21:12:08+0000 waiting for reconnect for 4.9938740730285645 more seconds
2023-05-12T21:12:08+0000 waiting for reconnect for 2.9927923679351807 more seconds
2023-05-12T21:12:09+0000 waiting for reconnect for 6.997904062271118 more seconds
2023-05-12T21:12:09+0000 waiting for reconnect for 8.998390197753906 more seconds
2023-05-12T21:12:09+0000 waiting for reconnect for 0.9915776252746582 more seconds
2023-05-12T21:12:09+0000 waiting for reconnect for 3.992349624633789 more seconds
2023-05-12T21:12:09+0000 waiting for reconnect for 1.9910469055175781 more seconds
2023-05-12T21:12:10+0000 waiting for reconnect for 5.996617317199707 more seconds
2023-05-12T21:12:10+0000 waiting for reconnect for 7.997318506240845 more seconds
2023-05-12T21:12:10+0000 stopping reactor
2023-05-12T21:12:10+0000 Main loop terminated.
```

Any ideas on how to improve this on the user end? Can we improve the error message there? And perhaps revisit the documentation on crossbar configuration, `ocsbow crossbar generate_config` is still documented there and seems like its deprecated:
```
$ ocsbow crossbar generate_config
usage: ocsbow [-h] [--site SITE] [--site-file SITE_FILE] [--site-host SITE_HOST] [--site-hub SITE_HUB] [--site-http SITE_HTTP] [--site-realm SITE_REALM]
              [--instance-id INSTANCE_ID] [--address-root ADDRESS_ROOT] [--registry-address REGISTRY_ADDRESS] [--log-dir LOG_DIR] [--working-dir WORKING_DIR]
              {status,up,down,config} ...
ocsbow: error: argument command: invalid choice: 'crossbar' (choose from 'status', 'up', 'down', 'config')
```

And the newer `ocs-local-support generate_crossbar_config` seems to not work either:
```
$ ocs-local-support generate_crossbar_config
  File "/home/koopman/git/ocs/.venv/bin/ocs-local-support", line 8, in <module>
    sys.exit(main_local())
  File "/home/koopman/git/ocs/.venv/lib/python3.10/site-packages/ocs/ocsbow.py", line 1014, in main_local
    cm = supports.crossbar['ctrl']
KeyError: 'ctrl'
```

BrianJKoopman (commented):
This is great! However now we have to deal with users configuring the crossbar server, which we avoided originally, bundling a sensible default in the crossbar Docker image we provide (with the `observatory.` URI.) I think our documentation on this needs some updating.

docs/user/site_config.rst
BrianJKoopman: Maybe we can build an entrypoint script for the crossbar container that allows for setting the address root in the crossbar container's environment variables? Then users would have to have matching address roots in the SCF and the docker-compose file for their crossbar server, which would look like:
```
  crossbar:
    image: simonsobs/ocs-crossbar:latest
    ports:
      - "127.0.0.1:8001:8001"
    environment:
         - PYTHONUNBUFFERED=1
         - ADDRESS_ROOT=lab1
```

mhasself: Ok, see what I've done in latest commit.  Happy to change envvar names, etc.

I updated the docs, minimally, but I think:
- We should drop the part about rebuilding images with modified config.json ... an advanced docker user would already know how to do that, and bind-mounting is good enough for most OCS users.
- We should move the docker part to be before the generate_crossbar_config stuff.

mhasself:
Thanks -- I think I've addressed all your suggestions now.  The connection error now presents a more detailed error message, with advice if possible.

ocs/agents/influxdb_publisher/drivers.py
BrianJKoopman: That's alright, just wanted to make sure I understood what was going on. Thanks for confirming!

docs/user/site_config.rst
BrianJKoopman: Looks great! Thanks for the updates.

I agree on both those points. The link in the "Example Config" section is going to break now too. Since `ocs/docker/crossbar/config.json` and `ocs/ocs/support/crossbar_config.json` are the same now, I'd change some of the wording in that section, and maybe just "include" the file on that page for viewing, rather than link to it.

Do you want to make those docs changes, or shall I? Happy to help there.

mhasself: > Do you want to make those docs changes, or shall I? Happy to help there.

Cool!  If you could take over from here I think that would be the most efficient way forward.

BrianJKoopman (approved):
Alright, I've made some modifications to the documentation. Will merge when checks pass.

PR #332, opened 2023-05-22
Cast tuple param before type check

### LINKED ISSUE #274: param decorator doen't support type check on tuples
For context, this came up in socs on the ACU Agent.

When using the param decorator, if trying to type check a tuple, i.e.:
```
@ocs_agent.param('azpts', type=tuple)
```

Before the check occurs the input parameter value is converted to and from json, effectively casting it to a list. This results in a failed type check, for example:
```
>>> resp = client.constant_velocity_scan(azpts=(20., 120.), el=35., azvel=2., acc=2., ntimes=8, azonly=False, simulator=True)
>>> print(resp)
OCSReply: ERROR : Param 'azpts'=[20.0, 120.0] is not of required type (<class 'tuple'>)
```

We should check if the type is `tuple` and cast before the check within this if block: https://github.com/simonsobs/ocs/blob/076105e9613ef78745d37df3d9bf8a38ce1e28bd/ocs/ocs_agent.py#L1221

BrianJKoopman:
## Description

This PR adds tuple support to the `@ocs_agent.params` decorator.

Tuples inadvertently get cast to lists when converted to json as they are sent from the client to the agent, this just casts back to tuple when `type=tuple`.

## Motivation and Context

Fixes #274. Issue originally came up when using the ACU Agent in socs.

## How Has This Been Tested?

Only tested in in the unit tests. But I made the test return exactly the error in #274 before applying the fix.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/ocs_agent.py
+++ b/ocs/ocs_agent.py
@@ -1389,6 +1389,9 @@ class ParamHandler:
                 # Free cast from int to float.
                 if type is float and isinstance(value, int):
                     value = float(value)
+                # Fix type after json conversion
+                if type is tuple and isinstance(value, list) and cast in [tuple, None]:
+                    value = tuple(value)
                 if not isinstance(value, type):
                     raise ParamError(f"Param '{key}'={value} is not of required type ({type})")
             if choices is not None:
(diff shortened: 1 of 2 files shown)


mhasself (approved):
I guess this works!  Another solution is to say "tuples aren't allowed".

PR #333, opened 2023-05-23
Support direct start calls with params to param decorated Tasks/Processes

### LINKED ISSUE #251: Calling a param decorated operation directly with no params results in hidden error
If an Operation is decorated with `@ocs_agent.param` and is called directly with `OCSAgent.start()`, as might be done within an Agent (i.e. in the [Lakeshore 372 Agent](https://github.com/simonsobs/socs/blob/5798dcf1e7c3f86eb726905460bbf193f3feb03d/agents/lakeshore372/LS372_agent.py#L208)), but without any parameters, then the `ParamHandler` is going to fail and an `AttributeError` will be caught.

However, this error doesn't propagate to the client, because there is no associated session (which returns on OK status.) To the Client it looks like the operation has never run, and no indication of a problem is made in the Agent logs, they only reflect that the start call was made.

This was encountered when developing the Lakeshore425 Agent: https://github.com/simonsobs/socs/pull/225

I think we should do a few things:
1. Log the caught exceptions within `start()` in the Agent logs
2. Add a test for `ParamHandler` for this scenario
3. Handle params being `None` within `ParamHandler`, taking the defaults set in the decorator

The ParamHandler runs into issues on this line: https://github.com/simonsobs/ocs/blob/aa6fda03e9e1e82f0c318e15d6d5f48c9b550836/ocs/ocs_agent.py#L1208

**OCS Version:** 0.9.0+10

BrianJKoopman:
## Description

This PR fixes a bug detailed in #251. The summary is that if a start call is made within an Agent and not passed any params (i.e. `params=None`) then the call fails silently.

Added to resolve this:
* Logging of caught exceptions when calling `self.agent.start()`. This provides some indication that the start call failed within the Agent logs (that is, the agent making the start call, not necessarily the Agent receiving the start.)
* A test for task start that replicates the error.
* Handling of `params=None` where we take the default value provided in the `@ocs_agent.param` decorator.

## Motivation and Context

Resolves #251.

Direct calls like this are being used in a couple of active agent PRs in socs, including the HWP agents. Would be nice to finally fix this here.

## How Has This Been Tested?

I added a small task to the fake data agent locally that makes a start call to the acq process without passing an params, which reproduced the bug in #251. I then added the logging for the caught exceptions.

Then I added a test that would fail given the existing behavior, before implementing a fix. Once the fix was in place I ran the modified fake data agent task again and verified the acq process ran as expected with the default value from the param decorator.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/ocs_agent.py
+++ b/ocs/ocs_agent.py
@@ -1304,6 +1304,8 @@ class ParamHandler:
     """
 
     def __init__(self, params):
+        if params is None:
+            params = {}
         self._params = params
         self._checked = set()
 
@@ -1374,11 +1376,7 @@ class ParamHandler:
 
         """
         self._checked.add(key)
-        if self._params is None and not isinstance(default, ParamError):
-            self._params = {}  # avoid attribute error in check_for_strays
-            value = default
-        else:
-            value = self._params.get(key, None)
+        value = self._params.get(key, None)
         is_unset = value is None and \
             (treat_none_as_missing or key not in self._params)
         if is_unset:
--- a/ocs/ocs_agent.py
+++ b/ocs/ocs_agent.py
@@ -694,8 +694,10 @@ class OCSAgent(ApplicationSession):
                     handler = ParamHandler(params)
                     params = handler.batch(op.launcher._ocs_prescreen)
                 except ParamError as err:
+                    self.log.error(f"Caught ParamError during start call: {err}")
                     return (ocs.ERROR, err.msg, {})
                 except Exception as err:
+                    self.log.error(f"Caught Exception during start call: {err}")
                     return (ocs.ERROR, f'CRASH: during param pre-processing: {str(err)}', {})
 
             # Mark as started.
--- a/ocs/ocs_agent.py
+++ b/ocs/ocs_agent.py
@@ -694,10 +694,10 @@ class OCSAgent(ApplicationSession):
                     handler = ParamHandler(params)
                     params = handler.batch(op.launcher._ocs_prescreen)
                 except ParamError as err:
-                    self.log.error(f"Caught ParamError during start call: {err}")
+                    self.log.error("Caught ParamError during start call: {err}", err=err)
                     return (ocs.ERROR, err.msg, {})
                 except Exception as err:
-                    self.log.error(f"Caught Exception during start call: {err}")
+                    self.log.error("Caught Exception during start call: {err}", err=err)
                     return (ocs.ERROR, f'CRASH: during param pre-processing: {str(err)}', {})
 
             # Mark as started.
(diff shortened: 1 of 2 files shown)


ocs/ocs_agent.py
```diff
@@ -1372,7 +1374,11 @@ def get(self, key, default=ParamError(''), check=None, cast=None, type=None,
 
         """
         self._checked.add(key)
-        value = self._params.get(key, None)
+        if self._params is None and not isinstance(default, ParamError):
+            self._params = {}  # avoid attribute error in check_for_strays
+            value = default
+        else:
+            value = self._params.get(key, None)
```
mhasself, line 1381: Is it perhaps simpler to detect and convert None -> {} in the `__init__`?

BrianJKoopman: Mhm, when I considered that I dropped the `self._params is None` in line 1377, but then detecting when to use the default value seemed tricky. But I didn't realize at the time you were handling that in the lines below...

New implementation in [1a4257c](https://github.com/simonsobs/ocs/pull/333/commits/1a4257c89254b1826bc56d1da18f1f8456db409e).

ocs/ocs_agent.py
```diff
@@ -694,8 +694,10 @@ def start(self, op_name, params=None):
                     handler = ParamHandler(params)
                     params = handler.batch(op.launcher._ocs_prescreen)
                 except ParamError as err:
+                    self.log.error(f"Caught ParamError during start call: {err}")
                     return (ocs.ERROR, err.msg, {})
                 except Exception as err:
+                    self.log.error(f"Caught Exception during start call: {err}")
```
mhasself, line 700: Passing a string containing random repr(exception) to log.error seems to almost always lead to a crash, because repr(err) has some {} in it, and log.error tries to string interpolate it.  I believe
```
self.log.error("Caught Exception during start call: {err}", err=err)
```
is how you're supposed to do it.

mhasself (approved):
See one concern.

ocs/ocs_agent.py
BrianJKoopman: Mhm, I remember seeing that crash often a while back, but not so much these days, I wonder if something changed. I did test the f-string evaluation (see below) and it was working, but I've change to use the standard txaio syntax anyway.

```
>>> import txaio
>>> txaio.use_twisted()
>>> txaio.start_logging()
>>> log = txaio.make_logger()
>>> log.info('test')
2023-05-24T10:27:40-0400 test
>>> log.info('test {foo}', foo='bar')
2023-05-24T10:29:51-0400 test bar
>>> foo='bar'
>>> log.info(f'test {foo}')
2023-05-24T10:28:20-0400 test bar
```

BrianJKoopman: Ahh, I see, that's a helpful example. Thanks!

Issue #367, opened 2024-01-05
Race condition in wait()


mhasself:
See discussion in https://github.com/simonsobs/socs/issues/607

Issue #357, opened 2023-10-05
Processes should be able to self-report as "degraded"


mhasself:
Currently we monitor agent instance health through each process' "OpCode", which mostly boils down to whether the process is running or not.  ([See enum](https://github.com/simonsobs/ocs/blob/f6cd025930136781d12ad6b1bd9f50117969fe4f/ocs/base.py#L38).)  This value is included in agent heartbeat; is monitored and streamed to a feed by the registry.

In the model where processes are long-lived, and they deal with hardware problems by simply attempting reconnection, the process running/not-running is not sufficiently informative.  Such a process could record the status in session.data somehow ... but for generic propagation of that information we should standardize how that is done and how to get that information out.

Continuing from the discussion on dev call yesterday, a proposal is:
- Add "DEGRADED" to the OpCode enum.
- Standardize on `session.data["degraded_at"] = unix_timestamp` to mark running sessions as degraded.
- In OpSession, add function "set_degraded(degraded [bool])" to mark / clear the degraded state (which just updates session.data).
- In OpSession.op_code property, if status == "running" but data['degraded'] > 0 then return value "DEGRADED".

Individual agents and processes will need to manually implement the use of degraded, if it is applicable to how they deal with errors.  Alarms configured for such agents will need to be updated to map the "degraded" state to be as bad as "not running".

mhasself:
While we're in there, might be nice to have processes automagically transition out of "starting" state when the process code is run, rather than relying on agent code to set_status('running') manually.

mhasself:
Comment from @jlashner -- maybe session.degraded instead of using session.data['degraded'], as some agents might be storing a non-dict in session.data, which would break session.data.get.

However: don't do that ... can we find all agents that mis-use session.data?

mhasself:
Addressed in #371.

PR #371, opened 2024-01-08
Improvements to session.status

mhasself:
## Description

Primary changes:
- `session.op_code` enum has a new value "DEGRADED" (8), which is returned in the case that an operation is "running" but has `session.data['degraded']` exists and is True
- `session.status` will automatically be put in state "running" just before launching the agent operation code
- When running `session.set_status` from a "blocking" operation (i.e. from a worker thread), the code now blocks while waiting for the status to change (the change is made in the reactor thread).  This means that one can reasonably expect that `set_status('running')` does not return until `session.status == 'running'`.  (This is less useful, given the point above ...
- `session.publish_status` has been removed. Previously, this was called whenever `session.add_message` was called, and caused the encoded session to be pushed to `{agent_address}.feed`.  I am pretty sure that no one uses this feed.  The `Registry` uses the heartbeat to get op_code.

As a result, also:
- All agents in ocs are all updated to not call session.set_status() unnecessarily.
- OCS docs updated to not call session.set_status() unnecessarily.
- FakeDataAgent acq process has options to exercise the DEGRADED state.

The side-effects of this should be minimal.  Agents that call `session.set_status('running')`, even though it's already running, will not get an error or anything.  Agents that don't implement session.data['degraded'] don't need to worry about activating the DEGRADED op code.

## Motivation and Context

Addresses #357 and #367.

## How Has This Been Tested?

Tested extensively with FakeDataAgent.  Before removing interfaces, I checked that there are no users in ocs, socs, ocs-web.

## Types of changes

## Checklist:

### DIFF
--- a/docs/developer/clients.rst
+++ b/docs/developer/clients.rst
@@ -216,6 +216,32 @@ of a Task::
     else:
         print('Task did not complete successfully')
 
+Check Health of a Process
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Usually it's good when a Process is still running, but some Processes
+will also mark themselves as "degraded" if there is some non-fatal
+problem that means the acquisition or control is not occurring::
+
+    from ocs.ocs_client import OCSClient
+
+    client = OCSClient('agent-instance-id')
+
+    client.some_process.start()
+    time.sleep(2)
+    status = client.some_process()
+
+    if response.session['success'] is None:
+        if response.session['degraded']:
+            print('Process is running, but in a degraded state.')
+        else:
+            print('Process is running (and does not report degraded state).')
+    elif response.session['success']:
+        print('Process exited without error')
+    else:
+        print('Process exited with error')
+
+
 Check Latest Data in an Operation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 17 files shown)


mhasself:
Just a warning that during PR revisions I will be force-pushing this branch to change my commit message typo "have-baked" to the intended "half-baked" :P

mhasself:
Comment from @jlashner -- maybe session.degraded instead of using session.data['degraded'], as some agents might be storing a non-dict in session.data, which would break session.data.get.

However: don't do that ... can we find all agents that mis-use session.data?

mhasself:
Latest commit changes to using `session.degraded` instead of `session.data['degraded']`.

I suppose this needs something in the docs, too; perhaps someone can suggest exactly where that would best be added.

BrianJKoopman (approved):
This looks good to me!

> Latest commit changes to using `session.degraded` instead of `session.data['degraded']`.
> 
> I suppose this needs something in the docs, too; perhaps someone can suggest exactly where that would best be added.

How about an example for how to check for degraded status on "Clients and Control Programs" page?

https://ocs.readthedocs.io/en/main/developer/clients.html#examples

mhasself:
> How about an example for how to check for degraded status on "Clients and Control Programs" page?
> 
> https://ocs.readthedocs.io/en/main/developer/clients.html#examples

Thanks.  Added an example to the docs.  Will rebase when docs build is fixed on main.

BrianJKoopman:
Docs build now fixed on `main`, feel free to merge/rebase whenever.

PR #383, opened 2024-04-25
HostManager: use "docker compose" instead of "docker-compose"

### LINKED ISSUE #382: Add Docker Compose v2 support to HostManager
Compose v1 support [ended in June 2023](https://www.docker.com/blog/new-docker-compose-v2-and-v1-deprecation/). We still use it, but we should add support for Compose v2 to the HostManager for newer systems where it might be more awkward to install Compose v1.

Compose v2 is built into Docker now and is run as `docker compose` rather than `docker-compose` for v1. Perhaps we just add a check for the presence of v2 and prefer its use if it's available? (I should note in most places we're using v1 right now v2 is actually also available, so that will cause a switch there. If necessary we add a flag for compose version preference?)

Motivated by #380, and the fact that we're nearly a year beyond the v1 support date.

mhasself:
## Description

When managing docker containers, instead of using a mixture of calls to `docker` and `docker-compose` CLIs, the code now exclusively uses `docker` (through the "v2" interface `docker compose ...`).

This work also removes the "--docker-compose-bin" argument, which was never fully propagated, in any case (especially "up" and "down" were always run with whatever docker-compose was in the PATH); and "inspect" was (and continues to be) run with the default docker binary in PATH.

## Motivation and Context

Resolves #382 

## How Has This Been Tested?

Tested with a bunch of FakeDataAgents on system running Docker version 25.0.2.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/agents/host_manager/agent.py
+++ b/ocs/agents/host_manager/agent.py
@@ -27,13 +27,12 @@ class HostManager:
     that host (either automatically or on request).
     """
 
-    def __init__(self, agent, docker_composes=[], docker_compose_bin=None,
+    def __init__(self, agent, docker_composes=[],
                  docker_service_prefix='ocs-'):
         self.agent = agent
         self.running = False
         self.database = {}  # key is instance_id (or docker service name).
         self.docker_composes = docker_composes
-        self.docker_compose_bin = docker_compose_bin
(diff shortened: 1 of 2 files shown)


BrianJKoopman (approved):
Looks great, thanks for the fast turnaround!

PR #387, opened 2024-05-30
Adds message_idx to OCS feeds to protect against duplicate data in the aggregator

jlashner:
## Description

This PR adds the field `message_idx` to the OCSFeed object, which is included in the encoded feed data sent with each message. This is incremented every time a message is sent, and is checked in the aggregator agent to prevent saving duplicate data packets, which we've seen a number of times (f[or instance here](https://github.com/simonsobs/daq-discussions/discussions/100))

## Motivation and Context

This should solve issues we've seen with duplicate samples in the aggregator, as is seen here: https://github.com/simonsobs/daq-discussions/discussions/100

## How Has This Been Tested?

This has not yet been tested.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/agents/aggregator/drivers.py
+++ b/ocs/agents/aggregator/drivers.py
@@ -174,6 +174,7 @@ class Provider:
         self.frame_length = frame_length
         self.prov_id = prov_id
         self.log = txaio.make_logger()
+        self.last_message_idx = None
 
         self.blocks = {}
 
@@ -315,7 +316,7 @@ class Provider:
 
         return new_data
 
-    def save_to_block(self, data):
+    def save_to_block(self, data, message_idx=None):
         """Saves a list of data points into blocks. A block will be created
         for any new block_name.
 
@@ -335,9 +336,16 @@ class Provider:
             These must match -- in this instance both the word 'test'.
 
         Args:
-            data (dict): data dictionary from incoming data queue
-
+            data (dict):
+                data dictionary from incoming data queue
+            message_idx (int, optional):
+                message index for the incoming data. If a message index is repeated, the data will be ignored.
         """
+        if message_idx is not None:
+            if message_idx == self.last_message_idx: # Ignore duplicate messages
+                return
+            self.last_message_idx = message_idx
+
         self.refresh()
 
         if self.frame_start_time is None:
@@ -597,7 +605,8 @@ class Aggregator:
                 pid = self.add_provider(address, sessid, **prov_kwargs)
 
             prov = self.providers[pid]
-            prov.save_to_block(data)
+            message_idx = feed.get('message_idx', None)
+            prov.save_to_block(data, message_idx=message_idx)
 
     def add_provider(self, prov_address, prov_sessid, **prov_kwargs):
         """
--- a/ocs/agents/aggregator/drivers.py
+++ b/ocs/agents/aggregator/drivers.py
@@ -342,7 +342,7 @@ class Provider:
                 message index for the incoming data. If a message index is repeated, the data will be ignored.
         """
         if message_idx is not None:
-            if message_idx == self.last_message_idx: # Ignore duplicate messages
+            if message_idx == self.last_message_idx:  # Ignore duplicate messages
                 return
             self.last_message_idx = message_idx
 
--- a/ocs/ocs_feed.py
+++ b/ocs/ocs_feed.py
@@ -242,6 +242,7 @@ class Feed:
             # Publish message immediately
             try:
                 self.agent.publish(self.address, (message, self.encoded()))
+                self.message_idx += 1
             except TransportLost:
                 self.agent.log.error('Could not publish to Feed. TransportLost. '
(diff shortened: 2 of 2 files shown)


ocs/ocs_feed.py
```diff
@@ -136,6 +138,7 @@ def flush_buffer(self):
                     {k: b.encoded() for k, b in self.blocks.items() if b.timestamps},
                     self.encoded()
                 ))
+                self.message_idx += 1
```
mhasself, line 141: This should probably not be inside the try.  The blocks are going to get cleared, so you'll want to bump message_idx.

ocs/ocs_feed.py
```diff
@@ -239,6 +242,7 @@ def publish_message(self, message, timestamp=None):
             # Publish message immediately
             try:
                 self.agent.publish(self.address, (message, self.encoded()))
+                self.message_idx += 1
```
mhasself, line 245: Same here; move out of try.

Consider making "increment_message_idx" an optional parameter of `self.encoded()`?

ocs/agents/aggregator/drivers.py
```diff
@@ -335,9 +336,16 @@ def save_to_block(self, data):
...
         Args:
-            data (dict): data dictionary from incoming data queue
-
+            data (dict):
+                data dictionary from incoming data queue
+            message_idx (int, optional):
+                message index for the incoming data. If a message index is repeated, the data will be ignored.
         """
+        if message_idx is not None:
+            if message_idx == self.last_message_idx:  # Ignore duplicate messages
```
mhasself, line 345: Consider `<=` here?

That could lead to out-of-order things getting lost.  But if things are out of order, this duplicate protection won't work anyway.

I don't think things can arrive out of order though.

ocs/agents/aggregator/drivers.py
```diff
@@ -335,9 +336,16 @@ def save_to_block(self, data):
...
+                return
```
mhasself, line 346: I think we'd still like to log when this happens.  Maybe in a summary way ("dropped x duplicate frames in the past 1 hour") to not swamp the logs.  But _something_ to help us understand that this happening.

ocs/agents/aggregator/drivers.py
```diff
@@ -597,7 +605,8 @@ def process_incoming_data(self):
                 pid = self.add_provider(address, sessid, **prov_kwargs)
 
             prov = self.providers[pid]
-            prov.save_to_block(data)
+            message_idx = feed.get('message_idx', None)
```
mhasself, line 608: Why is the processing of message_idx in save_to_block instead of here?

mhasself (changes requested):
Some thoughts inline... probably we'll want a test for this, right?

ocs/ocs_feed.py
jlashner: My intention was for the `message_idx` to increment for each successfully published message, rather than for each attempted message, but am open to changing this if there is a good reason.

I believe there's not a huge difference between the two modes... The only difference I can think of now is that if the receiver observes an index skip, they can be confident that its not an issue on the sender's side, and that it's a network / crossbar problem.

jlashner: I guess the counterpoint is that if we move it out of the try then the receiver can know whether they're missing data that the agent intended to send, possibly due to agent-side issues.

mhasself: > My intention was for the `message_idx` to increment for each successfully published message, rather than for each attempted message, but am open to changing this if there is a good reason.
> 
> I believe there's not a huge difference between the two modes... The only difference I can think of now is that if the receiver observes an index skip, they can be confident that its not an issue on the sender's side, and that it's a network / crossbar problem.

But is a failure at this point in the agent a guarantee that the message won't make it to a consumer?  I know that's _likely_ the case, but I'm doubtful that it's guaranteed.  It is much safer to never risk sending the same message_idx out on the wire.

jlashner: ok ya I see, so the increment should probably happen before the try statement, or in a finally block

jlashner: or like you say, after the encoding. Having the increment happen in the encoding function also could be right, but it feels weird to me since I don't love having the encode function change the feed state, even if its behind an optional kwarg.

jlashner:
Should we close this PR given that we found that the issue is with the bookbinder and not data being aggregated twice? (https://github.com/simonsobs/daq-discussions/discussions/100#discussioncomment-9655510)

BrianJKoopman:
> Should we close this PR given that we found that the issue is with the bookbinder and not data being aggregated twice? ([simonsobs/daq-discussions#100 (comment)](https://github.com/simonsobs/daq-discussions/discussions/100#discussioncomment-9655510))

Yeah, I think that's fine. It sounds like the problem would have persisted with this anyway. We can come back to this in the future if we think it's necessary. Thanks for working on it though!

Issue #391, opened 2024-07-29
ocs container builds are failing due to bug in setuptools


BrianJKoopman:
Builds currently fail with the following error:
```
$ docker build --no-cache -t ocs .

...snip....

Step 14/16 : RUN pip3 install .
 ---> Running in 41f49f20402f
Processing /app/ocs
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /usr/bin/python3 /tmp/tmpx1r3y0c0 get_requires_for_build_wheel /tmp/tmpya52bdh4
       cwd: /tmp/pip-req-build-8uhb4nj9
  Complete output (49 lines):
  Warning: build in /tmp/pip-build-env-a5hyznc_/overlay/lib/python3.8/site-packages is using versioneer.py from /tmp/pip-req-build-8uhb4nj9/versioneer.py
  running egg_info
  writing ocs.egg-info/PKG-INFO
  writing dependency_links to ocs.egg-info/dependency_links.txt
  writing entry points to ocs.egg-info/entry_points.txt
  writing requirements to ocs.egg-info/requires.txt
  writing top-level names to ocs.egg-info/top_level.txt
  reading manifest file 'ocs.egg-info/SOURCES.txt'
  reading manifest template 'MANIFEST.in'
  adding license file 'LICENSE.txt'
  Traceback (most recent call last):
    File "/tmp/tmpx1r3y0c0", line 280, in <module>
      main()
    File "/tmp/tmpx1r3y0c0", line 263, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/tmp/tmpx1r3y0c0", line 114, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-a5hyznc_/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 327, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=[])
    File "/tmp/pip-build-env-a5hyznc_/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 297, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-a5hyznc_/overlay/lib/python3.8/site-packages/setuptools/build_meta.py", line 313, in run_setup
      exec(code, locals())
[...]

PR #403, opened 2024-09-03
Switch build-backend from setuptools to hatchling

BrianJKoopman:
## Description

This PR switches the build system backend from `setuptools` to `hatchling`.

Since `versioneer` only supports `setuptools` based projects, we drop that as well, removing the vendored copy we were including. The closest functionality I could find to that was through `hatch-vcs`. This results in slightly different versions compared to `versioneer`, illustrated by this example:

**versioneer**: `0.11.2+7.ga1e7e91`
**hatch-vcs**: `0.11.2.post1.dev7+ga1e7e91`

In switching backends we entirely replace `setup.py`/`setup.cfg` with the more modern `pyproject.toml`.

## Motivation and Context

When `setuptools` broke our builds (see #391) I wanted to explore modernizing our builds by migrating to a newer backend. There are many choices, but I settled on `hatchling` because some other SO projects (like nextline) use it, and it's the first tab in [documentation in the Python Packaging User Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/), so seems inherently recommended.

## How Has This Been Tested?

I have built the package, installed locally, and run the tests on this installation. I've also inspected the contents of the tarball compared to the latest version (v0.11.2) up on PyPI. I've dropped including the `tests/` directory in the sdist. `setuptools` was only partially picking this up. I debated including `tests/` and `docs/` in the sdist, but have left them out at this time. Let me know if you feel we should include them.

## Types of changes

- [x] Packaging change

## Checklist:


BrianJKoopman:
Going ahead with this after discussion in the meeting today.

PR #411, opened 2024-10-21
Setup coverage reporting on codecov.io

BrianJKoopman:
## Description

This PR sets up code coverage reporting on codecov.io. Before merging it might even remove coverage reporting to coveralls.

## Motivation and Context

We've been using coveralls.io for a while now, but I often find the coverage reports are extremely slow to navigate, taking 10's of seconds or longer to expand directories/load source files. I'd like to migrate away to codecov.io for this reason alone, but codecov has some additional nice features that I find appealing. codecov is also used by several other repos in our organization.

## How Has This Been Tested?

I've migrated [sorunlib](https://github.com/simonsobs/sorunlib) to codecov already, so this is based on that.

## Types of changes

- [x] Repository infrastructure change

## Checklist:

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,14 @@
+---
+# https://docs.codecov.com/docs/pull-request-comments#disable-comment
+comment: false
+
+# https://docs.codecov.com/docs/commit-status
+coverage:
+  status:
+    project:
+      default:
+        target: auto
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
For reports to start working this needs to be merged. Replacement of coveralls will come in a follow-up PR.

PR #418, opened 2024-12-16
Adds kill_on_exit to _AgentRunner in testing

jlashner:
## Description

Adds the option to kill agents in tests using SIGKILL instead of SIGINT.

## Motivation and Context

In my development of of the HWP emulator-based integration tests (https://github.com/simonsobs/socs/pull/802) I'm finding that, though the test are working, shutting agents down with a single interrupt command sometimes does not work, and is fairly slow (~10 sec per agent for the agents that do work). I _believe_ this could be something related how and when socket connections are closed by the emulator but I'm not entirely sure.

I have found that using a SIGKILL instead of SIGINT to shutdown agents is much faster and more reliable, and I don't really think we care about clean shutdowns at the end of states, so I think this is a good option for some tests, such as the HWP integration tests which use many different agents.

## How Has This Been Tested?

This has been tested by running HWP integration tests from the PR above locally.

## Types of changes

## Checklist:

### DIFF
--- a/ocs/testing.py
+++ b/ocs/testing.py
@@ -23,8 +23,8 @@ class _AgentRunner:
             i.e. '../agents/fake_data/fake_data_agent.py'
         agent_name (str): Short, unique name for the agent
         args (list): Additional CLI arguments to add when starting the Agent
-        kill_to_exit (bool): If True, will send a kill signal to exit instead of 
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
Looks good to me!

PR #407, opened 2024-09-19
Remove old agent plugin system

BrianJKoopman:
## Description

This PR removes the old plugin system from its remaining usage in `ocsbow` and the `HostManagerAgent`. I also removed the docs for the plugin scan section of the [site config page](https://ocs.readthedocs.io/en/main/developer/site_config.html#agent-script-discovery), and any references to 'agent-paths' in the SCF, which aren't used by the new plugin system.

## Motivation and Context

The new plugin system has been around for just about two years now, introduced in [v0.10.0](https://github.com/simonsobs/ocs/releases/tag/v0.10.0). In `socs` we've continued to maintain both [ocs_plugin_so.py](https://github.com/simonsobs/socs/blob/main/socs/agents/`ocs_plugin_so.py`) (the old plugin file) alongside the new `plugin.py` and I'd like to stop doing that.

Resolves #322.

## How Has This Been Tested?

I've run a small ocs network with two Host Managers, one for docker and one for agents running from the new plugin system. I've also run `ocsbow config plugins` and confirmed output. The output changed a little bit, just based on how the two plugin systems worked. The old system would show full paths to the modules like this:
```
$ ocsbow config plugins
ocs plugin detection
--------------------
Scanning.
  ... adding to path: /home/ocs/git/ocs/ocs/agents/
  ... adding to path: /home/ocs/git/socs/socs/agents/
Found:
  RegistryAgent        : /home/ocs/git/ocs/ocs/agents/registry/agent.py
  AggregatorAgent      : /home/ocs/git/ocs/ocs/agents/aggregator/agent.py
  HostManager          : /home/ocs/git/ocs/ocs/agents/host_manager/agent.py
  FakeDataAgent        : /home/ocs/git/ocs/ocs/agents/fake_data/agent.py
  InfluxDBAgent        : /home/ocs/git/ocs/ocs/agents/influxdb_publisher/agent.py
  BarebonesAgent       : /home/ocs/git/ocs/ocs/agents/barebones/agent.py
  ACUAgent             : /home/ocs/git/socs/socs/agents/acu/agent.py
  BlueforsAgent        : /home/ocs/git/socs/socs/agents/bluefors/agent.py
  CrateAgent           : /home/ocs/git/socs/socs/agents/smurf_crate_monitor/agent.py
  CryomechCPAAgent     : /home/ocs/git/socs/socs/agents/cryomech_cpa/agent.py
  FPGAAgent            : /home/ocs/git/socs/socs/agents/holo_fpga/agent.py
[...]

PR #420, opened 2025-01-17
Remove long deprecated `client_t.run_control_script2`

BrianJKoopman:
## Description

This PR removes `client_t.run_control_script2`, which we deprecated a little over 4 years ago now.

## Motivation and Context

We haven't really used the client interface provided by `client_t` ever since `matched_client`/`ocs_client` was introduced.

Also, with the removal of the old plugin system in #407, I'd like the next release of ocs to include some clean up and remove interfaces we've had deprecated for a while. Be on the look out for similar PRs for other deprecated components soon.

## How Has This Been Tested?

Hasn't really been tested. I tried running the miniobs example, but ran into some issues before even getting to the `run_control_script` part...

## Types of changes

- [x] Code cleanup

## Checklist:

### DIFF
--- a/docs/developer/site_config.rst
+++ b/docs/developer/site_config.rst
@@ -169,36 +169,3 @@ additional site-info settable from the site's parser.
 .. note::
     For examples calling these commandline arguments see
     :ref:`ocs_agent_cmdline_examples`.
-
-
-Control Clients and the Site Config
-===================================
-
-As of this writing, Control Clients do not store configuration in the
-SCF.  But there is an interim interface available for Control Clients
-to access the Site Configuration, with the usual command-line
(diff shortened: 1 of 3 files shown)


mhasself (approved):
Cool.  Ya if you need to delete miniobs, I understand.  I haven't used it in a while I don't think.

BrianJKoopman:
> Cool. Ya if you need to delete miniobs, I understand. I haven't used it in a while I don't think.

Thanks, that'll likely happen. crossbar had some dependency issues on install, and then sort of ran, but threw some errors. Didn't spend much time on it, but with crossbar being as it is it's probably not worth the hassle.

PR #459, opened 2026-03-24
influxdb: Add session.data with connection status

### LINKED ISSUE #326: influxdb needs session.data
If the influxdb server stops, for example, there's no way to tell that in ocs-web or any other ocs client.

The "record" process should report the influxdb connection state (and maybe some other stats!).

BrianJKoopman:
## Description

This PR adds a `session.data` object for the InfluxDB Publisher agents. This contains just the InfluxDB connection state and a timestamp at the moment, but can easily be expanded.

The implementation differs a bit between the v1 and v2 agents, but connection status was added to the driver's `Publisher` class as an attribute, and the connection state is used to set the session as 'degraded' or not.

## Motivation and Context

Resolves #326.

## How Has This Been Tested?

I have tested this with a local test setup that includes InfluxDB, the Publisher agents, and other supporting ocs infrastructure like crossbar. I've checked the session output:
```
In [2]: client.record.status().session['data']
Out[2]: {'connected': True, 'last_updated': 1774389203.53926}
```

And the degraded state:
```
In [7]: client.record.status().session['degraded']
Out[7]: True

In [8]: client.record.status().session['degraded']
Out[8]: False
```

## Types of changes

## Checklist:

### DIFF
--- a/ocs/agents/influxdb_publisher_v2/agent.py
+++ b/ocs/agents/influxdb_publisher_v2/agent.py
@@ -83,6 +83,13 @@ class InfluxDBAgentv2:
             test_mode (bool, optional): Run the record Process loop only once.
                 This is meant only for testing. Default is False.
 
+        Notes:
+            An example of the session data::
+
+                >>> response.session['data']
+                {'connected': True,
+                 'last_updated': 1774389203.53926}
+
         """
         self.aggregate = True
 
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
Going to go ahead with this merge, as I'm building the new tagging features on top of this branch, and I'm happy with its functionality.

Issue #175, opened 2020-09-30
Improve tagging in Influx Publisher Agent


BrianJKoopman:
To improve the ease of plotting in Grafana it was suggested we could improve the tagging of our measurements in the InfluxDB. This should be explored and expanded.

PR #466, opened 2026-04-07
influxdb: Refactor common drivers to prepare for tagging

BrianJKoopman:
## Description

This PR refactors the common InfluxDB drivers in order to make it easier to add InfluxDB tagging for #175.

## Motivation and Context

This code is fairly old and kind of a mess. It's maybe still kind of messy, but at least a messy that made it easier to add tagging.

## How Has This Been Tested?

The first commit (f56539f4d30c5e64f2de41b7d5d616594688f6f9) here adds a test for the 'json' protocol, which was missing. And 98a825ad69789b121c48b0695bcfd847be1d4f9b adds more tests for single 'timestamp' style messages, which were also missing.

Also, throughout development I had the old code still running with an `assert` statement checking the new code output the same structure. I ran the agent locally using both the 'json' and 'line' protocols.

## Types of changes

- [x] Code refactor (non-breaking changes which change the structure of the code, but have no behavioral changes)

## Checklist:

### DIFF
--- a/ocs/common/influxdb_drivers.py
+++ b/ocs/common/influxdb_drivers.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from datetime import datetime, timezone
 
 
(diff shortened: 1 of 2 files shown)


mhasself (approved):
Seems like a good refactoring --

perhaps not the time, but reading this made me wonder about how invalid data are handled.  Generally log messages like "Warning: Cannot convert {timestamp} to an InfluxDB compatible time." are not super helpful... it's desirable to know what agent is producing weird results.  Raising an error instead of silently dropping things _could_ be handled more informatively by the wrapping code.

BrianJKoopman:
> Seems like a good refactoring --
> 
> perhaps not the time, but reading this made me wonder about how invalid data are handled. Generally log messages like "Warning: Cannot convert {timestamp} to an InfluxDB compatible time." are not super helpful... it's desirable to know what agent is producing weird results. Raising an error instead of silently dropping things _could_ be handled more informatively by the wrapping code.

Thanks for the review. Good point about the error text there not being particularly informative. I should have had you review #451, where I added that initially.

This was handling a pretty rare edge case that came up when one of the new ACUs was generating strange timestamps. I agree the wrapping code could raise a more informative error message about which agent sent the invalid data. I'll file a separate issue and merge for now.

We might even consider having the agent side validate the data before sending too, since we can more easily inform the user that something is wrong by crashing the agent, much like we do if a block changes structure, or [a field name is invalid](https://github.com/simonsobs/ocs/blob/4199ac1474318692a06884e24395715d5838e2e6/ocs/ocs_feed.py#L211-L214), etc.

